PR #5, opened 2018-03-22
fakestorage: add support for Crc32c checksum of object content.

solidsystem:
Hi, This change makes it possible to test client code that is using Crc32c checksum checks.

Thanks for creating this server by the way, works great for my tests!

### DIFF
--- a/fakestorage/object_test.go
+++ b/fakestorage/object_test.go
@@ -10,10 +10,11 @@ import (
 	"reflect"
 	"testing"
 
-	"cloud.google.com/go/storage"
 	"encoding/binary"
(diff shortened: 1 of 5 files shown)


fakestorage/object_test.go
```diff
@@ -10,16 +10,32 @@ import (
 	"reflect"
 	"testing"
 
+	"encoding/binary"
+	"hash/crc32"
```
fsouza: Can you move those up in the group of stdlib imports?

solidsystem:
Believe the imports are fixed now. Please let me know if I should look at anything else.

PR #6, opened 2018-04-13
Add filesystem backend

gracenoah:
This PR adds the ability to persist stored files to disk. This is helpful when using this server in development (as opposed to just in tests). Usage looks like:

```go
fakestorage.NewServerWithOptions(fakestorage.Options{
		InitialObjects: []fakestorage.Object{
			{
				BucketName: "bucket",
				Name:       "object",
				Content:    []byte("An example object."),
			},
		},
		Host:        "0.0.0.0",
		Port:        4443,
		StorageRoot: "/storage",
	})
```

This PR is broken up into 3 parts for easy reviewing:
- Separate the storage details from the server implementation, including basic smoke tests for the storage backend
- Implement an alternative filesystem implementation of the storage backend for persistence
- Clean ups and better error handling.

I took special care to make sure I don't break the existing package API, but because some methods didn't have an error return argument, I had to add some panics. I think that's a reasonable tradeoff for now and the errors can be returned when you decide it's okay to break backwards compatibility.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -18,17 +18,19 @@ import (
 func (s *Server) CreateBucket(name string) {
 	s.mtx.Lock()
 	defer s.mtx.Unlock()
-	if _, ok := s.buckets[name]; !ok {
-		s.buckets[name] = nil
+	err := s.backend.CreateBucket(name)
+	if err != nil {
+		panic(err)
 	}
 }
 
 func (s *Server) listBuckets(w http.ResponseWriter, r *http.Request) {
 	s.mtx.RLock()
 	defer s.mtx.RUnlock()
-	bucketNames := make([]string, 0, len(s.buckets))
-	for name := range s.buckets {
-		bucketNames = append(bucketNames, name)
+
+	bucketNames, err := s.backend.ListBuckets()
+	if err != nil {
+		panic(err)
 	}
 	resp := newListBucketsResponse(bucketNames)
 	json.NewEncoder(w).Encode(resp)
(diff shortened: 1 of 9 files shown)


solidsystem:
A couple of comments, since I see you are touching upon the Crc32c checksum functionality I added.

It looks like you remove the Crc32c property from the Object struct, and calculate it on the fly when returning the object instead. I assume this is because it simplifies the filesystem backend implementation, the Object can then just be a normal file with a name and content. 

One reason I think it might be beneficial to keep the property on the struct is being able to easily test the failure mode of clients, e.g. injecting a wrong checksum and letting the server return it to test client-side error handling. 

My other slight objection to this is that I think I will want to add support for metadata on the object in the future, just like the real cloud storage server has. This should be relatively easy with the current in-memory object structure, but would require adding the necessary metadata solution to this new filesystem backend solution, just like I guess you want to avoid doing now by removing the Crc32c property on the Object struct.

Just my two cents. I think it will probably also be useful to have a filesystem backend, so I don't want to stop that from happening.

gracenoah:
Sounds reasonable. Considering how hopelessly unsafe and inefficient the current implementation is (because every request is written directly to the file non-atomically and the entire objects are stored in memory most of the time), what do you think about just storing the backend.Object json-encoded into the file? The alternative I was considering was to store two files - one for the contents and one for the metadata, but that feel likes a premature optimization and added complexity for something that's not meant to be efficient or robust.

solidsystem:
Yeah, I was thinking about the same actually, just storing the json encoded Object struct seems like a good solution to me.

PR #18, opened 2019-03-14
fakestorage: add support for MD5 checksum of object content

whiteley:
We'd like to use this library to test some GCS related code that relies on the MD5 object attributes. From my testing, I believe this to be correct and I greatly cribbed off the crc32c implementation. Happy to implement any improvements suggested though.

Thanks for the library!


fsouza:
@whiteley thanks for contributing!

PR #23, opened 2019-03-27
Using lowercase keys in bucketResponse struct json

teone:
Google Cloud Storage REST APIs use lowercase keys in json bucket response.
This will get the fake server to work with the GCS library.
Without this `client.list_buckets()` is returning:
```
[<Bucket: None>, <Bucket: None>]
```
with this patch:
```
[<Bucket: another-bucket>, <Bucket: bucket>]
```


fsouza:
@teone thanks for contributing!

PR #28, opened 2019-04-19
Adding download route to the mux server

teone:
Hi @fsouza,
here is a small change to get the fake server to run together with the python client library.

I have a somehow working version of the library packaged in a container here: https://github.com/teone/gc-fake-storage


fsouza:
@teone nice! Thanks for contributing, I didn't know about the ``/download`` endpoint x)

Nice to see the docker image too. I can add a note to the readme about it, if you want!

snawaz:
The GCS official doc does not seem to support any endpoint with starting `/download`. It does support the regular pattern with `alt=media` (query-string) to download the object:

https://cloud.google.com/storage/docs/downloading-objects

The upload API does have `/upload` endpoint though.

teone:
@snawaz if I recall correctly that endpoint was called by the python library (that I was using as client at the time).

PR #39, opened 2019-06-11
Update uploadFileContent for Python client

jwhitlock:
This updates ``uploadFileContent`` to handle the needs of both the Go and Python clients when uploading over multiple requests. It rewrites ``parseRange`` as ``parseContentRange``, to parse the header content and leave interpretation and setting response headers to ``uploadFileContent``.

The Go client always sends content as one or more chunks without a total size, and, when the stream is exhausted, sends a final request with no content but the final size.

The Go client also sends the header ``X-Guploader-No-308``. This specifies that the response to a "mid-stream" request should be sent with a status code ``200 OK`` plus the header ``X-Http-Status-Code-Override: 308``.

The Python client sends chunks with a total size when known, such as when uploading a file or an in-memory string. Unlike a streaming response, the final request contains content. and the final request contains data. It expects a mid-stream response to have a status code ``308 Permanent Redirect``.

The Python client is pickier about some response details, and ``uploadFileContent``'s responses are changed to work with both the Python and (more flexible) Go clients:

 * The Python client requires the ``Range`` response header to start counting at 0, such as ``bytes 0-999``, while the Go client appears to ignore the start byte.
 * The Python client requires that the final response is a ``200 OK``, while the Go client accepts this or a ``201 Created``.

### DIFF
--- a/fakestorage/upload_test.go
+++ b/fakestorage/upload_test.go
@@ -221,6 +221,7 @@ func TestServerInvalidUploadType(t *testing.T) {
 }
 
 func TestParseContentRange(t *testing.T) {
+	t.Parallel()
 	var goodHeaderTests = []struct {
 		header string
 		output contentRange
@@ -242,6 +243,7 @@ func TestParseContentRange(t *testing.T) {
 	for _, test := range goodHeaderTests {
 		test := test
 		t.Run(test.header, func(t *testing.T) {
+			t.Parallel()
 			output, err := parseContentRange(test.header)
 			if output != test.output {
 				t.Fatalf("output is different.\nexpected: %+v\n  actual: %+v\n", test.output, output)
@@ -264,6 +266,7 @@ func TestParseContentRange(t *testing.T) {
 	for _, test := range badHeaderTests {
 		test := test
 		t.Run(test, func(t *testing.T) {
+			t.Parallel()
 			_, err := parseContentRange(test)
 			if err == nil {
 				t.Fatalf("Expected err!=<nil>, but was %v", err)
--- a/fakestorage/upload_test.go
+++ b/fakestorage/upload_test.go
@@ -219,3 +219,55 @@ func TestServerInvalidUploadType(t *testing.T) {
 		t.Errorf("wrong status returned\nwant %d\ngot  %d", expectedStatus, resp.StatusCode)
 	}
 }
+
+func TestParseContentRange(t *testing.T) {
+	var goodHeaderTests = []struct {
+		header string
+		output contentRange
+	}{
+		{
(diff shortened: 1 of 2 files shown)


fakestorage/upload_test.go
```diff
@@ -219,3 +219,55 @@ func TestServerInvalidUploadType(t *testing.T) {
 		t.Errorf("wrong status returned\nwant %d\ngot  %d", expectedStatus, resp.StatusCode)
 	}
 }
+
+func TestParseContentRange(t *testing.T) {
+	var goodHeaderTests = []struct {
```
fsouza: Minor, but can you mark this test (and its sub-tests) as parallel with ``t.Parallel()``?

jwhitlock:
Thanks for your quick reviews and feedback on these PRs. It's increased my confidence in using this codebase, even though Python is my primary language.

>Also, I wonder if we should add some kind of integration test to make sure we remain compatible with the Python client (or if that would be more in the context of gc-fake-storage). What do you think?

I've thought about this a lot. My thought are probably too many for a now closed PR 😁 . An issue may be appropriate to coordinate on next steps.

I've been running integration tests in https://github.com/jwhitlock/test-fake-gcs, and it has been very useful for rapidly testing and developing code. I also had to fork https://github.com/teone/gc-fake-storage as https://github.com/jwhitlock/gc-fake-storage/tree/test, because I needed to test the latest code, and my PRs have gone unanswered for two weeks.

My preference would be for this repo to expand to cover gc-fake-storage and test-fake-gcs. It would provide the Go library, build and publish a docker image of the server, host the sample code to extend the official GCS libraries, and provide a docker-compose configuration for integration testing with those modified client libraries. This could start with Python and Go, and grow to others as contributors discover the project. Having all the pieces in one place would allow for rapid development and coordination. It also may make it a candidate for Google to adopt it some day as an official emulator.

However, you may see benefits to the current split between fake-gcs-server the library and gc-fake-storage the server / docker image. If teone/gc-fake-storage is a dead project, then I would be interested in getting the replacement repo started.

fsouza:
@jwhitlock hey, apologies for the delayed response. I think that bringing gc-fake-storage in makes sense (as ``cmd/gcs-emulator`` maybe? or ``fake-gcs``? something like that, not sure). Then if we have that, we pack it up in a Docker image and have some kind of downstream repo that runs integration tests using SDKs, or even bringing the integration tests into this repo too.

@teone any thoughts?

teone:
@fsouza I'd be totally happy to merge `fake-gcs` with these repo and have a single codebase! Next couple week is going to be crazy at work so I won't have time to help on this, but I can dedicate some time to this the second half of July

PR #40, opened 2019-06-12
Add optional custom public host

jwhitlock:
Add the ability to specify a different public host than ``storage.googleapis.com``. This would be used by https://github.com/teone/gc-fake-storage to serve files.  This is set before ``buildMuxer()`` is called, and can be returned by ``PublicURL()``.


fsouza:
@jwhitlock thank you very much!

PR #46, opened 2019-07-16
cmd: add gcs-emulator

### LINKED ISSUE #45: Add a dockerized server configuration
https://github.com/teone/gc-fake-storage provides a ``Dockerfile`` and related scripts to allow this code to be run as a server, so that it can be used outside of tests of Go code. This Docker image is published at https://hub.docker.com/r/matteoscandolo/gc-fake-storage.

It would be easier to improve and test this use case if the Docker configuration was part of this project instead. It could be easier to coordinate creating a new image when the library changes. Also, contributors could collaborate on language-specific clients, verified by integration tests.

There is no explicit license on https://github.com/teone/gc-fake-storage, which I believe means @teone has the copyright, which makes it incompatible with this project's BSD 2 license, copyright Francisco Souza. I believe the clearest way to merge the code is if @teone makes the initial PR, which makes it clear he's contributing under the license for this this project. After the initial merge of files, I'd be happy to make additional changes. 
 
This was first discussed on closed PR https://github.com/fsouza/fake-gcs-server/pull/39#issuecomment-501852822

fsouza:
Initial iteration/proposal of the gcs-emulator command (which may be renamed to something ``fake-gcs`` or something like that). It includes a dockerfile and integration with Drone (building docker images in travis is more painful than doing it from drone).

Closes #45.

cc @teone @jwhitlock


.drone.yml
```diff
@@ -0,0 +1,199 @@
...
+  image: golang:1.13beta1
+  commands:
+  - go build -o gcs-emulator -mod readonly ./cmd/gcs-emulator
+  depends_on:
+  - mod-download
+
+- name: test-ci-dockerfile
+  image: plugins/docker
+  settings:
+    dockerfile: drone/Dockerfile
```
jwhitlock: I'm not sure how Drone works, but I tried this:

```
# mod-download?
go mod download
# build?
go build -o gcs-emulator -mod readonly ./cmd/gcs-emulator
# test-ci-dockerfile?
docker build . --file drone/Dockerfile --tag plugins/docker:test
# Run it?
docker run plugins/docker:test
```
This results in:
```
standard_init_linux.go:207: exec user process caused "exec format error"
```

The docker image created from the root ``Dockerfile`` does appear to do the right thing.

fsouza: @jwhitlock I assume you're on a Mac? `go build -o gcs-emulator -mod readonly ./cmd/gcs-emulator` generates a Mach-O binary runnable on Mac and Docker runs on Linux, so the command would need to be something like ``GOOS=linux go build -o gcs-emulator -mod readonly ./cmd/gcs-emulator`` to get go to cross compile it (drone doesn't need to set GOOS=linux because it's also running on Linux, see https://cloud.drone.io/fsouza/fake-gcs-server/20/1/6).

But let's move the conversation to #47, as that seems to be a more complete solution! I can move it from travis to drone independently or push the changes there.

jwhitlock: Good guess! Switching to ``GOOS=linux go build -o gcs-emulator -mod readonly ./cmd/gcs-emulator`` make this work for me.

fsouza:
Closing in favor of #47.

PR #47, opened 2019-07-16
Adding dockerfile and code to run as development or integration test server

### LINKED ISSUE #45: Add a dockerized server configuration

teone:
Here is the first of the contribution discussed in #45 

Comments are welcome (also let me know if the file structure make sense to you!)

Closes #45


README.md
```diff
@@ -4,7 +4,74 @@
...
+
+## Example with the `python` client library:
+
+For a more complex example look into the `examples` directory
+
+```python
+# virtualenv venv --no-site-packages
+# source venv/bin/activate
+# pip install -r pip_requirements.txt
+# python python.py
```
jwhitlock: I think it would be easier to add more clients if we namespace by language, like ``examples/python/requirements.txt`` and ``examples/python/example.py``

README.md
```diff
@@ -4,7 +4,74 @@
...
+
+```
+- sample_bucket
+  - some_file.txt
+```
+
+To make sure everything works as expected you can execut these commands:
+
+```shell
+curl --insecure https://0.0.0.0:4443/storage/v1/b
```
jwhitlock: 👍 these sample commands worked for me

main.go
```diff
@@ -0,0 +1,86 @@
...
+
+		for _, f := range files {
+			fn := f.Name()
+			fi, err := os.Stat("/data/" + fn)
+
+			if err != nil {
+				panic(err)
+			}
+
+			switch mode := fi.Mode(); {
```
jwhitlock: I think we should make ``gocritic`` happy and switch this to an ``if`` statement (see [TravisCI output](https://travis-ci.com/fsouza/fake-gcs-server/jobs/216685618#L327-L330)).

main.go
```diff
@@ -0,0 +1,86 @@
+package main
```
jwhitlock: I think the copyright transfer should be explicit with a header, in this and other files:

```
// Copyright 2019 Francisco Souza. All rights reserved.
// Use of this source code is governed by a BSD-style
// license that can be found in the LICENSE file.
```

VERSION
```diff
@@ -0,0 +1 @@
+1.7.2
```
jwhitlock: I like versioning the Docker image, and keeping the image tag in sync with the library version.

Dockerfile
```diff
@@ -0,0 +1,17 @@
+FROM golang:1.12-stretch as builder
+MAINTAINER Matteo Scandolo <teo.punto@gmail.com>
+
+WORKDIR /go
+ENV GO111MODULE=on
+ADD . /go/src/github.com/fsouza/fake-gcs-server
```
fsouza: Since we're using go modules, this doesn't need to be added to the GOPATH, so we could do something like:

```dockerfile
WORKDIR /code
ADD . ./
RUN CGO_ENABLED=0 GOOS=linux go build -o /build/go-fake-storage
```

We can probably set GOPROXY to ``https://proxy.golang.org`` to speed up the build as well.

Dockerfile
```diff
@@ -0,0 +1,17 @@
...
+RUN CGO_ENABLED=0 GOOS=linux go build -o /build/go-fake-storage /go/src/github.com/fsouza/fake-gcs-server/main.go
+
+FROM alpine:3.5
```
fsouza: Let's update this to 3.10.1 before merging.

```suggestion
FROM alpine:3.10.1
```

VERSION
fsouza: #46 addressed that with an option in Drone: https://github.com/fsouza/fake-gcs-server/pull/46/files#diff-3216dfff0ed3e301453e6799e8c367e2R71

We can probably reuse it and get rid of the ``VERSION`` file (having the VERSION file gets tricky as we'd need to remember to always update it before tagging, with a CI it's automatic).

(I can do the drone integration part)

README.md
fsouza: +1 for namespacing!

teone:
@fsouza I should have some time too this week, so let address the comments on copyright, `switch` statement and the `example`, if that's ok I'll leave to you the decision on the versioning strategies (the version file is just something I'm used to, but we have a CI job to make sure we don't mess it up).

I'll take a look at `GOPROXY`, even though this looks to be [another interesting option](https://medium.com/@pliutau/docker-and-go-modules-4265894f9fc) to speed up the build

fsouza:
@teone that option is helpful for subsequent builds, but won't help in the CI :(

I did have it in #46, though: https://github.com/fsouza/fake-gcs-server/pull/46/files#diff-3254677a7917c6c01f55212f86c57fbfR7

.gitignore
```diff
@@ -1,2 +1,6 @@
 Gopkg.lock
 vendor
+
+# IDE Specifics
+
+.idea
```
fsouza: Do you mind not putting IDE specifics here? It gets tricky to track different editors x)

(As a tip for your local environment, you can add ``.idea`` to a [global .gitignore file](https://help.github.com/en/articles/ignoring-files#create-a-global-gitignore)!)

main.go
```diff
@@ -0,0 +1,90 @@
...
+	"os"
+
+	"github.com/fsouza/fake-gcs-server/fakestorage"
+)
+
+func generateObjectsFromFiles() []fakestorage.Object {
+	objects := []fakestorage.Object{}
+
+	// if a /data volume is mounted in the container
+	if _, err := os.Stat("/data"); !os.IsNotExist(err) {
```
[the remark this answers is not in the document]
jwhitlock: I liked the config refactor of PR #46, making the paths parameters could be part of that near future update. It may be cleaned to skip the initial load if the path is not defined.

fsouza:
Cool, will merge this and wire-up the build system! :tada:

Issue #53, opened 2019-08-03
Trying to download blob calls original googleapis url


kye308:
Hi, thanks for all the great work so far! I'm interested in downloading files with the server and am not sure if that is supported right now.

I can follow the steps in the readme to get a blob object. I would like to call `blob.download_as_string()` (or any similar download function) but when I do, I get 

```
Unauthorized: 401 GET https://www.googleapis.com/download/storage/v1/b/<bucket>/o<object_name>?alt=media: (u'Request failed with status code', 401, u'Expected one of', 200, 206)
```
Is this expected? Wasn't sure if blob download is supported atm.

fsouza:
Hi @kye308 can you try the example from the [examples](https://github.com/fsouza/fake-gcs-server/tree/master/examples/python) folder?

fsouza:
I just replaced the snippet with a direct link. The script is also being tested in our CI environment now, so we should be good!

Please feel free to reopen if you still encounter this issue.

PR #54, opened 2019-08-06
Python example - set the public host

jwhitlock:
Change ``BASE_URL`` to ``EXTERNAL_URL``, to mirror the setting in the ``fakestorage`` server. Add ``PUBLIC_HOST`` as well, to allow coordinating the public download host with the ``fakestorage`` server.

This may be the problem behind issues #53. I've tested similar but more complicated code in https://github.com/jwhitlock/test-fake-gcs.

### DIFF
--- a/examples/python/python.py
+++ b/examples/python/python.py
@@ -7,16 +7,18 @@ import urllib3
 from google.auth.credentials import AnonymousCredentials
 from google.cloud import storage
 
-BASE_URL = "https://127.0.0.1:4443"
+EXTERNAL_URL = "https://127.0.0.1:4443"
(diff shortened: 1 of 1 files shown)


fsouza:
@jwhitlock thanks for this change. It looks like #53 mentions the example in the readme, should we just link the full code from the README instead of having it inline there? Sounds easier than syncing up the code in the examples folder with the snippet in the readme (I can do it in a separate change)

fsouza:
Merging. I'm also going to add that test script to the CI.

jwhitlock:
I agree, duplication should be avoided. Thanks for updating the README to link to the example.

PR #60, opened 2019-08-13
Initial object versioning support - retention of old generations if version is enabled, and getObject by generation

dcaba:
Finally the first version of the versioning support is ready for review. First I will describe what has been covered, a functionality I want to add really soon (object listing that considers archived versions) and the ones I dont expect to cover in short term (I don't need them yet, and I think is not critical functionality to enjoy versioning), and I will add inline comments to highlight some decisions I've made that probably are worth to bear in mind during the review.

What we have covered:
* Related to versioning implementation:
   * We have targeted the in memory backend, extending the backend interface: fs backend will not store the new fields, and will refuse/fail on any operation that targets versioning or specific generations
   * Object storing updated/created/deleted dates, we can create or just inherit from inputs (but delete time, that will be always consistent)
   * Return consistent ObjectAttrs.Generation
   * Bucket versioning flag (BucketAttrs.VersioningEnabled) available for the in memory engine, and modifying the CreateBucket function (with a mandatory parameter, versioning enabled or not), and also implying a bucket is no longer just the bucketName, but an object with several fields.
      * Bucket listing / props in bucket level responses considering the versioning status
   * Keeping multiple object versions when bucket versioning is enabled, returning always the last
   * Implement deletion with bucket versioning is enabled, returning consistent results (archived vs live versions)
   * Implement lookups of specific generations, that can target archived version
   * Returning a consistent object ID (that should contain name, bucket and generation)
* Non related to the functionality, but accelerates testing:
   * Execute linting+tests as part of the Dockerbuild 
      * This may be rolled back depending on the owner point of view: we are replicating, and sadly duplicating, logic we have in CI to reproduce tests locally

In the next PR we expect to cover:
* Implement object listing with Query.Versions = true

Things we expect not cover under this PR neither short term (if the previous implementation does not require to proceed in the end)
* Deletion of specific generations / archived objects
* Adding support to Conditions (in order to perform conditional operations as described [here](https://godoc.org/cloud.google.com/go/storage#hdr-Conditions) ), specifically GenerationMatch
* Bucket metageneration management
* Object metageneration management
* lifecycles rules
* Support of versions in the filesystem backend (that can be fun...)

Thanks for your patience, thanks so much for your project and hope this does not generate a big headache.


dcaba:
Hi @fsouza ! 
 
  still WIP (but maybe there are some parts I should split and just raise as different PRs?), but if you have any opinions around this work, feel free to let me know.

   Best!

fsouza:
Hey @dcaba, this sounds amazing! This week is kinda busy, but I'll try to have a look on this later today or over the weekend!

Thank you very much for opening it!

dcaba:
> Hey @dcaba, this sounds amazing! This week is kinda busy, but I'll try to have a look on this later today or over the weekend!
> 
> Thank you very much for opening it!

No rush @fsouza , and apologies: it's still WIP. Hope I can continue (and hopefully finish) next week, but yeah, if you want to start taking a look, great.

   Best!

dcaba:
Hi @fsouza : these days are being more challenging than expected in my side, but i hope i can continue next week. And yeah, happy to start getting comments for any part you dont like! I added a note in the PR description describing a bit the plans: in regards to conditions, `Conditions.GenerationMatch` is probably something I will need to cover, but probably not everything... so, yeah, i guess there will be parts to extend after my contribution

fakestorage/object.go
```diff
@@ -137,6 +174,15 @@ func (s *Server) GetObject(bucketName, objectName string) (Object, error) {
 	return obj, nil
 }
 
+func (s *Server) GetObjectWithGeneration(bucketName, objectName string, generation int64) (Object, error) {
```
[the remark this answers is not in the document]
dcaba: was already addressed

internal/backend/object.go
```diff
@@ -13,9 +17,17 @@ type Object struct {
...
+	Updated     string
+	Generation  int64
 }
 
 // ID is useful for comparing objects
 func (o *Object) ID() string {
-	return o.BucketName + "/" + o.Name
+	return fmt.Sprintf("%s#%d", o.IDNoGen(), o.Generation)
+}
+// IDNoGen does not consider the generation field
```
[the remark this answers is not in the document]
dcaba: I dont see the difference... maybe refers to an old commit

fakestorage/bucket.go
```diff
@@ -14,9 +14,9 @@ import (
 // CreateBucket creates a bucket inside the server, so any API calls that
 // require the bucket name will recognize this bucket.
 //
-// If the bucket already exists, this method does nothing.
-func (s *Server) CreateBucket(name string) {
-	err := s.backend.CreateBucket(name)
+// If the bucket already exists, this method does nothing but panics if props differs
+func (s *Server) CreateBucket(name string, versioningEnabled bool) {
```
dcaba: This change is disruptive: we extend the previous contract with users of fakestorage requiring an extra arg (versioning enabled or not). If we dont want to introduce this breaking change, we can expose a CreateBucketWithVersioning, or something like CreateBucketWithOpts and the new arg (or accepting a struct that describes the Opts, that makes that more extensible), making the old (that can be marked as deprecated?) to call to the new one

fakestorage/response.go
```diff
@@ -19,30 +20,36 @@ type listResponse struct {
...
 }
 
-func newListBucketsResponse(bucketNames []string) listResponse {
+func newListBucketsResponse(buckets []backend.Bucket) listResponse {
 	resp := listResponse{
 		Kind:  "storage#buckets",
-		Items: make([]interface{}, len(bucketNames)),
+		Items: make([]interface{}, len(buckets)),
 	}
-	sort.Strings(bucketNames)
```
dcaba: we were previously always replying to bucket listings with ordered lists... but i'd say that does not match gcs real behavior, and despite it may facilitate assertions, i think is better to return disordered lists / matching real behavior

fakestorage/object.go
```diff
@@ -105,6 +119,10 @@ func toBackendObjects(objects []Object) []backend.Object {
 			Crc32c:      o.Crc32c,
 			Md5Hash:     o.Md5Hash,
 			ACL:         string(o.ACL),
+			Created:     getCurrentIfZero(o.Created).Format(time.RFC3339),
```
dcaba: It was easier/natural to keep delete updated in the backend rather than here, but the backend is "silly" regarding creation/update times. We can move that there too if you think is convenient

internal/backend/backend_test.go
```diff
@@ -43,16 +43,48 @@ func testForStorageBackends(t *testing.T, test func(t *testing.T, storage Storag
 	}
 }
 
-func noError(t *testing.T, err error) {
+func noError(t *testing.T, err error, description string) {
```
dcaba: without description, it was difficult to track where errors came from

internal/backend/backend_test.go
```diff
@@ -62,95 +94,150 @@ func TestObjectCRUD(t *testing.T) {
...
-		if obj.BucketName != bucketName {
-			t.Errorf("wrong bucket name\nwant %q\ngot  %q", bucketName, obj.BucketName)
-		}
-		if obj.Name != objectName {
-			t.Errorf("wrong object name\n want %q\ngot  %q", objectName, obj.Name)
-		}
-		if !bytes.Equal(obj.Content, content2) {
-			t.Errorf("wrong object content\n want %q\ngot  %q", content2, obj.Content)
-		}
+	for _, versioningEnabled := range []bool{true, false} {
```
dcaba: This test maybe became too big, but very representative: CRUD but also considering versioning scenarios when that applies, so a complete flow of upload / overwrite / read initial object if versioning is enabled / delete / check delete acts as expected (versioning vs no versioning). We can try to break that down if needed

internal/backend/memory.go
```diff
@@ -8,117 +8,216 @@ import (
...
 // StorageMemory is an implementation of the backend storage that stores data in memory
 type StorageMemory struct {
-	buckets map[string][]Object
+	buckets map[string]bucketInMemory
 	mtx     sync.RWMutex
 }
 
+type bucketInMemory struct {
+	Bucket
+	activeObjects   []Object
```
dcaba: rather than lists, we could use map to speed up access... but i dont think this project needs to care too much about efficiency, so keeping lists of objects as we used to have, for simplicity

internal/backend/memory.go
```diff
@@ -8,117 +8,216 @@ import (
...
 )
 
 // StorageMemory is an implementation of the backend storage that stores data in memory
 type StorageMemory struct {
-	buckets map[string][]Object
+	buckets map[string]bucketInMemory
 	mtx     sync.RWMutex
 }
 
+type bucketInMemory struct {
```
dcaba: now, rather than a bucket being represented by a name (and implemented just as a list of objects associated to the name), we promote the Bucket entity (so props can be attached there) and the in-memory implementation extends the struct to store lists of objects.

dcaba:
@fsouza , ready for review. PR description updated, and added comments inline highlighting some implementation topics. Open to your feedback and extra changes. Just seen there has been recent changes conflicting with my changes: I will take a look asap, and work in a different branch from now to implement old-generations listing

dcaba:
> @fsouza , ready for review. PR description updated, and added comments inline highlighting some implementation topics. Open to your feedback and extra changes. Just seen there has been recent changes conflicting with my changes: I will take a look asap, and work in a different branch from now to implement old-generations listing

rebase done; no conflicts now

fakestorage/object_test.go
```diff
@@ -342,7 +475,133 @@ func TestServiceClientListObjects(t *testing.T) {
...
+				}
+				if !reflect.DeepEqual(prefixes, test.expectedPrefixes) {
+					t.Errorf("wrong prefixes returned\nwant %#v\ngot  %#v", test.expectedPrefixes, prefixes)
+				}
+			})
+		}
+	})
+}
+
+func TestServerClientListAfterCreateAndMultipleGenerations(t *testing.T) {
```
dcaba: this is going to be revisited in the next PR, because of the support of listing with generations

fsouza:
@dcaba this is awesome, I haven't had a chance to take another look at the PR yet, but I'd like to thank you for putting the work and getting this done! 🎉 🎉 🎉 

I'll likely be able to go through it tomorrow or Saturday.

fakestorage/object.go
```diff
@@ -107,6 +121,10 @@ func toBackendObjects(objects []Object) []backend.Object {
 			Crc32c:          o.Crc32c,
 			Md5Hash:         o.Md5Hash,
 			ACL:             o.ACL,
+			Created:         getCurrentIfZero(o.Created).Format(time.RFC3339),
```
dcaba: The backend storing dates as RFC3339 strings is something we can changed: I opted for this just to keep it aligned to the gcs API, but its true we are losing precision and we this can be decoupled

Dockerfile
```diff
@@ -2,15 +2,30 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-FROM golang:1.13.4-alpine AS builder
-
+FROM golang:1.13.4 AS tester
+ARG GOPROXY=https://proxy.golang.org
```
fsouza: This shouldn't be needed with Go 1.13, as there's now a default value for GOPROXY.

fakestorage/bucket.go
fsouza: Let's do ``CreateBucketWithOpts`` and deprecate this one. That way we can extend it in the future too, if needed.

Dockerfile
```diff
@@ -2,15 +2,30 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
-FROM golang:1.13.4-alpine AS builder
-
+FROM golang:1.13.4 AS tester
```
fsouza:
```suggestion
FROM golang:1.13.5 AS tester
```

it took me so long to review your PR that there's a new version of Go, sorry about that 🙊

Dockerfile
```diff
@@ -2,15 +2,30 @@
...
+
+FROM golangci/golangci-lint AS linter
+WORKDIR /code
+ADD . ./
+RUN golangci-lint run --enable-all \
+    -D errcheck -D lll -D dupl -D gochecknoglobals -D unparam \
+    --deadline 5m \
+    ./...
+
+FROM golang:1.13.4-alpine AS builder
```
fsouza:
```suggestion
FROM golang:1.13.5-alpine AS builder
```

same as above

fakestorage/object.go
```diff
@@ -124,11 +142,24 @@ func fromBackendObjects(objects []backend.Object) []Object {
...
 	return backendObjects
 }
 
+// https://github.com/googleapis/google-cloud-go/blob/2f857649c55302802e95b96119dd05032a61c87a/storage/storage.go#L1023
+func convertTimeWithoutError(t string) time.Time {
+	var r time.Time
+	if t != "" {
+		r, _ = time.Parse(time.RFC3339, t)
+	}
+	return r
```
fsouza: Can simplify this function:

```go
r, _ := time.Parse(time.RFC3339, t)
return r
```

internal/backend/fs.go
```diff
@@ -131,6 +137,11 @@ func (s *StorageFS) GetObject(bucketName, objectName string) (Object, error) {
 	return s.getObject(bucketName, objectName)
 }
 
+// GetObjectWithGeneration retrieves an specific version of the object. Not implemented
+func (s *StorageFS) GetObjectWithGeneration(bucketName, objectName string, generation int64) (Object, error) {
+	return Object{}, fmt.Errorf("not implemented: fs storage type does not support versioning yet")
```
fsouza:
```suggestion
	return Object{}, errors.New("not implemented: fs storage type does not support versioning yet")
```

internal/backend/fs.go
```diff
@@ -58,36 +61,39 @@ func (s *StorageFS) createBucket(name string) error {
...
 	defer s.mtx.RUnlock()
 	_, err := os.Stat(filepath.Join(s.rootDir, url.PathEscape(name)))
-	return err
+	return Bucket{Name: name}, err
 }
 
 // CreateObject stores an object
 func (s *StorageFS) CreateObject(obj Object) error {
+	if obj.Generation > 0 {
+		return fmt.Errorf("not implemented: fs storage type does not support objects generation yet")
```
fsouza:
```suggestion
		return errors.New("not implemented: fs storage type does not support objects generation yet")
```

fakestorage/object.go
fsouza: I think since the API will return RFC3339 anyways, it's fine to lose the precision in the backend.

fakestorage/object.go
```diff
@@ -164,7 +206,26 @@ func (s *Server) getObject(w http.ResponseWriter, r *http.Request) {
...
+	generationStr := r.FormValue("generation")
+	var (
+		obj        Object
+		err        error
+		generation int64
+	)
+	if generationStr != "" {
+		generation, err = strconv.ParseInt(generationStr, 10, 64)
+		if err != nil {
+			fmt.Println(err)
```
fsouza: Delete? 😁

fakestorage/object.go
```diff
@@ -258,7 +319,23 @@ func (s *Server) rewriteObject(w http.ResponseWriter, r *http.Request) {
...
+	var (
+		obj        Object
+		err        error
+		generation int64
+	)
+	generationStr := r.FormValue("generation")
+	if generationStr != "" {
+		generation, err = strconv.ParseInt(generationStr, 10, 64)
+		if err != nil {
+			fmt.Println(err)
```
fsouza: dd

fakestorage/response.go
fsouza: Makes sense. We can sort in the tests for assertions.

fakestorage/response.go
```diff
@@ -18,30 +19,36 @@ type listResponse struct {
...
-	return bucketResponse{
-		Kind: "storage#bucket",
-		ID:   bucketName,
-		Name: bucketName,
+type bucketVersioning struct {
+	Enabled bool `json:"enabled,omitempty"`
+}
+
+func newBucketResponse(bucketName string, versioningEnabled bool) (br bucketResponse) {
+	br = bucketResponse{
```
fsouza: Let's avoid the naked return.

```suggestion
	return bucketResponse{
```

fakestorage/response.go
```diff
@@ -18,30 +19,36 @@ type listResponse struct {
...
+		Kind:       "storage#bucket",
+		ID:         bucketName,
+		Name:       bucketName,
+		Versioning: &bucketVersioning{versioningEnabled},
 	}
+	return
```
fsouza:
```suggestion

```

internal/backend/memory.go
```diff
@@ -8,117 +8,216 @@ import (
...
+	if err != nil {
+		return Object{}, err
+	}
+	matchGeneration := false
+	obj := Object{BucketName: bucketName, Name: objectName}
+	listToConsider := bucketInMemory.activeObjects
+	if generation != 0 {
+		matchGeneration = true
+		obj.Generation = generation
+		listToConsider = append(listToConsider, bucketInMemory.archivedObjects...)
```
fsouza: I wonder if this can become too bad as ``archivedObjects`` grow. Not something to worry about now though :)

Hopefully it will never grow too much. We can adjust in the future if an issue is reported (or something like that)

internal/backend/backend_test.go
```diff
@@ -43,16 +43,48 @@ func testForStorageBackends(t *testing.T, test func(t *testing.T, storage Storag
...
+func noError(t *testing.T, err error, description string) {
 	if err != nil {
-		t.Fatal(err)
+		t.Fatalf("%s should not error, but got: %v", description, err)
 	}
 }
 
-func shouldError(t *testing.T, err error, message string) {
+func shouldError(t *testing.T, err error, description string) {
 	if err == nil {
```
fsouza: Should we mark this function as a test helper? This way it won't appear in the stack trace when the test fails.
```suggestion
	t.Helper()
	if err == nil {
```

internal/backend/backend_test.go
fsouza: Same as below. Should we mark this as a helper?

fakestorage/bucket.go
dcaba: Done, and updated the repo so we are using CreateBucketWithOpts everywhere. We could opt for a pattern a bit more sophisticated (https://github.com/tmrts/go-patterns/blob/master/idiom/functional-options.md), but at this stage, I think that may be enough. Is that fine for you?

Dockerfile
dcaba: Yeah! Rebased from master, that has this improvement too

Dockerfile
dcaba: No worries. I rebased from master, where you also updated

fakestorage/bucket.go
dcaba: (reopen if you have further comments - this and any other comment im marking as resolved, as im using them to track remaining work from my side)

internal/backend/backend_test.go
dcaba: Yes! In fact, i didnt know about the helper functionality until two weeks ago. I will incorporate, minimizing or removing the usage of descriptions.

fakestorage/object.go
dcaba: I used exactly the same function google is also using, probably not to couple the error response of time.Parse to the api response... but yeah, its quite natural that, in case of errors, they return a null value that is fine for us, so changing!

fakestorage/object.go
dcaba: good catch!

internal/backend/memory.go
dcaba: Its a very good point, but I think this is not a concern: looks the "unpack" with three dots does not actually extract each of the items and passes / manages them individually, but keeps the reference to the original archivedObjects slice (there are some interesting comments here: https://stackoverflow.com/questions/16248241/concatenate-two-slices-in-go ). In any case, an implementation with two searches, one in each slice, would be probably more performant (appending the slices has a cost in any case, as it may need to create a second slice to merge them), but in this specific scenario, I think its better to minimize duplication / keep current code readability rather than the performance win. Please, feel free to reopen if you think is worth changing that now

dcaba:
Hi @fsouza . Feedback has been considered. I have marked all comments as resolved (as ive been using them to track remaining work), but feel free to reopen or add extra comments if there is something you dont like. Thanks for your time!

Dockerfile
```diff
@@ -2,15 +2,31 @@
...
 ADD . ./
+RUN go test -race -vet all -mod readonly ./...
+
+FROM golangci/golangci-lint AS linter
+WORKDIR /code
+COPY --from=tester /go/pkg /go/pkg
+COPY --from=tester /code .
+RUN golangci-lint run --enable-all \
+    -D errcheck -D lll -D dupl -D gochecknoglobals -D unparam \
+    --deadline 5m \
```
dcaba: This has been a recent addition: i realized i was consuming a lot of space with docker images, and i realized this linter layer was storing more than 200M on each docker build after any code change... it was downloading all go packages again, and also golangci stores a cache to speed up future executions we will never reuse inside a docker layer. This has also improved performance

fsouza: Can you make it just ``golangci-lint run``? That way it uses the config specified in golangci.yml.

dcaba: Sorry, I didnt realize golangci-lint config file was introduced in master after my original branch (side effects of having this in my plate for so long :( ). I see you amended in master: I will start rebasing the subsequent PRs... i guess its going to take my entire life to resolve the conflicts :) News soon.

fsouza: You can `rebase -i` and drop the commits that were merged in this PR :D

dcaba: Thanks for the suggestion. Yeah... but what worries me are the commits post-review, which are not in the other branches yet, and given the incremental nature, probably are going to touch lines that other PRs are also modifying. In any case: the other PRs are quite small + (business as usual && life is hard), nothing to worry :)

PR #80, opened 2019-09-19
Add golang and nodejs examples

zigius:


examples/golang/main.go
```diff
@@ -0,0 +1,53 @@
...
+
+import (
+	"cloud.google.com/go/storage"
+	"context"
+	"crypto/tls"
+	"fmt"
+	"google.golang.org/api/iterator"
+	"google.golang.org/api/option"
+	"log"
+	"net/http"
```
fsouza: Can you run ``goimports`` to make sure the imports are grouped properly, or if you want to manually regroup them, just split the stdlib ones from the non-stdlib ones:

```go
import (
	"context"
	"crypto/tls"
	"fmt"
	"log"
	"net/http"

	"cloud.google.com/go/storage"
	"google.golang.org/api/iterator"
	"google.golang.org/api/option"
)
```

zigius: Done

PR #97, opened 2019-10-16
Support private acl for objects

lelikg:
allow to set private ACL on objects


fakestorage/response.go
```diff
@@ -4,7 +4,14 @@
 
 package fakestorage
 
-import "sort"
+import (
+	"net/http"
+	"sort"
+
+	cloudstorage "cloud.google.com/go/storage"
```
fsouza:
```suggestion
	"cloud.google.com/go/storage"
```

fakestorage/object.go
```diff
@@ -7,11 +7,14 @@ package fakestorage
...
 	"encoding/json"
 	"fmt"
+	"io/ioutil"
 	"net/http"
 	"sort"
 	"strconv"
 	"strings"
 
+	"cloud.google.com/go/storage"
+
```
fsouza: Please delete this line.

fakestorage/object.go
```diff
@@ -174,6 +180,48 @@ func (s *Server) deleteObject(w http.ResponseWriter, r *http.Request) {
...
+	bodyMap := make(map[string]string)
+	defer r.Body.Close()
+	body, err := ioutil.ReadAll(r.Body)
+	if err != nil {
+		http.Error(w, "bad request", http.StatusBadRequest)
+		return
+	}
+
+	json.Unmarshal(body, &bodyMap)
```
fsouza: Please check the error here. Also, where is bodyMap being used?

fakestorage/object_test.go
```diff
@@ -604,3 +604,42 @@ func TestServerClientObjectDeleteErrors(t *testing.T) {
...
+
+	runServersTest(t, objs, func(t *testing.T, server *Server) {
+		t.Run("public to private", func(t *testing.T) {
+			ctx := context.Background()
+			objHandle := server.Client().Bucket("some-bucket").Object("img/public-to-private.jpg")
+
+			err := objHandle.ACL().Set(ctx, storage.AllAuthenticatedUsers, storage.RoleReader)
+			if err != nil {
+				t.Errorf("unexpected error while setting acl %+v", err)
+				return
```
fsouza: use ``t.Fatalf`` instead of ``t.Errorf`` + ``return``.

fakestorage/object_test.go
```diff
@@ -604,3 +604,42 @@ func TestServerClientObjectDeleteErrors(t *testing.T) {
...
+			}
+
+			rules, err := objHandle.ACL().List(ctx)
+			if err != nil {
+				t.Errorf("unexpected error while getting acl %+v", err)
+				return
```
fsouza: same as above

fakestorage/object_test.go
```diff
@@ -604,3 +604,42 @@ func TestServerClientObjectDeleteErrors(t *testing.T) {
...
+			}
+
+			if len(rules) != 1 {
+				t.Error("acl has no rules")
+				return
```
fsouza: ditto

fakestorage/object_test.go
```diff
@@ -604,3 +604,42 @@ func TestServerClientObjectDeleteErrors(t *testing.T) {
...
+			}
+			if rules[0].Entity != storage.AllAuthenticatedUsers {
+				t.Error("acl entity not set to AllAuthenticatedUsers")
+				return
```
fsouza: ditto

fakestorage/object_test.go
```diff
@@ -604,3 +604,42 @@ func TestServerClientObjectDeleteErrors(t *testing.T) {
...
+			}
+
+			if rules[0].Role != storage.RoleReader {
+				t.Error("acl role not set to RoleReader")
+				return
```
fsouza: ditto

fakestorage/response.go
lelikg: this conflicts with "google.golang.org/api/storage/v1"

fakestorage/object.go
lelikg: I am using storage.ACLEntity in line 31

fakestorage/object.go
lelikg: removed

fakestorage/object_test.go
lelikg: done

fakestorage/object_test.go
lelikg: done

fakestorage/object_test.go
lelikg: done

fakestorage/object_test.go
lelikg: done

fakestorage/object_test.go
lelikg: done

lelikg:
> Looks mostly good! Wonder if we should add support for ACLs in other endpoints too (like object creation)?

IMHO it's not possible to set ACL on object creation, you can only do it after the object is already created.

fakestorage/object.go
fsouza: yeah I meant the blank line 🙈

lelikg: done

fsouza:
> > Looks mostly good! Wonder if we should add support for ACLs in other endpoints too (like object creation)?
> 
> IMHO it's not possible to set ACL on object creation, you can only do it after the object is already created.

What does this code do? Can we add a test to make sure it behaves as expected?

```go
package main

import (
	"context"

	"cloud.google.com/go/storage"
)

func main() {
	client, err := storage.NewClient(context.Background())
	if err != nil {
		panic(err)
	}
	obj := client.Bucket("some-bucket").Object("some/new-object.txt")
	writer := obj.NewWriter(context.Background())
	rule := storage.ACLRule{
		Entity: storage.AllUsers,
		Role:   storage.RoleReader,
	}
	writer.ObjectAttrs.ACL = append(writer.ObjectAttrs.ACL, rule)
	writer.Write([]byte("hi there"))
	err = writer.Close()
	if err != nil {
		panic(err)
	}
}
```

(I haven't tried running it, I can investigate later though)

lelikg:
> ```go
> storage.NewClient
> ```

Unfortunately the ACL is not written to the object when doing it this way

fsouza:
@lelikg sounds good. It looks like it can take PredefinedACL, but that's a different thing.

Thanks for contributing!

lelikg:
thanks for merging, is it possible to trigger a release?

fsouza:
Done, just pushed v1.13.0.

PR #108, opened 2019-10-24
multipartUpload accepts object name in query param

chrisrecalis:
As per documentation: https://cloud.google.com/storage/docs/json_api/v1/objects/insert the name of the object can either come from the query parameters or from the body.

It seems like its up to the implementors on which one to choose as the `go` storage client will send the name inside the body whereas the `nodejs` client sends it in the query params

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -146,9 +146,15 @@ func (s *Server) multipartUpload(bucketName string, w http.ResponseWriter, r *ht
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	objName := r.URL.Query().Get("name")
+	if objName == "" && metadata != nil {
+		objName = metadata.Name
+	}
+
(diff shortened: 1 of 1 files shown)


fakestorage/upload.go
```diff
@@ -146,9 +146,15 @@ func (s *Server) multipartUpload(bucketName string, w http.ResponseWriter, r *ht
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	objName := r.URL.Query().Get("name")
+	if objName == "" && metadata != nil {
```
fsouza: It's safe to assume metadata is not nil here.

```suggestion
	if objName == "" {
```

fsouza: Never mind, I take that back.

I'll add a test later to ensure we handle this properly x)

fsouza:
I'll merge once CI is green, thanks for contributing! 🎉

PR #109, opened 2019-10-24
github/workflows/main: compile golangci-lint on the fly

fsouza:
golangci-lint can be quickly installed using go modules, and that should
be faster. Without this change, our deps are getting downloaded twice.
While golangci-lint does have more dependencies than fake-gcs-server,
I'm banking on the fact that we won't need to download golangci-lint's
Docker image and this should speed things up :D

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -26,21 +26,26 @@ jobs:
 
       - uses: actions/checkout@v1.0.0
 
-      - name: go-mod-download
-        run: go mod download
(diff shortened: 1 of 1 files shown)


fsouza:
Comparison from a recent run:

- Pull golangci/golangci-lint: 26s
- install deps (go-mod-download): 14s
- run-linter: 43s

After this change:

- Pull golangci-golangci-lint: 0s 😁 
- install deps: 36s
- run-linter: 18s

PR #125, opened 2019-11-11
Support predefined canned ACL when creating object

lelikg:
also support full ACL rule instead of just entity

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -19,10 +19,11 @@ import (
 
 // Object represents the object that is stored within the fake server.
 type Object struct {
-	BucketName  string `json:"-"`
-	Name        string `json:"name"`
-	ContentType string `json:"contentType"`
-	Content     []byte `json:"-"`
+	BucketName      string `json:"-"`
+	Name            string `json:"name"`
+	ContentType     string `json:"contentType"`
+	ContentEncoding string `json:"contentEncoding"`
+	Content         []byte `json:"-"`
 	// Crc32c checksum of Content. calculated by server when it's upload methods are used.
 	Crc32c  string            `json:"crc32c,omitempty"`
 	Md5Hash string            `json:"md5hash,omitempty"`
@@ -99,13 +100,14 @@ func toBackendObjects(objects []Object) []backend.Object {
 	for _, o := range objects {
 		acl, _ := json.Marshal(o.ACL)
 		backendObjects = append(backendObjects, backend.Object{
-			BucketName:  o.BucketName,
-			Name:        o.Name,
-			Content:     o.Content,
-			ContentType: o.ContentType,
-			Crc32c:      o.Crc32c,
-			Md5Hash:     o.Md5Hash,
-			ACL:         acl,
+			BucketName:      o.BucketName,
+			Name:            o.Name,
+			Content:         o.Content,
+			ContentType:     o.ContentType,
+			ContentEncoding: o.ContentEncoding,
+			Crc32c:          o.Crc32c,
+			Md5Hash:         o.Md5Hash,
+			ACL:             acl,
 		})
 	}
 	return backendObjects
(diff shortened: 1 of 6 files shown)


lelikg:
Hi @fsouza, would be happy to merge this if no comments.

fsouza:
Hey @lelikg, thanks for opening this PR and sorry for the delayed response. I'll take a look at it later today!

fakestorage/object.go
```diff
@@ -19,14 +19,15 @@ import (
...
+	BucketName      string `json:"-"`
+	Name            string `json:"name"`
+	ContentType     string `json:"contentType"`
+	ContentEncoding string `json:"contentEncoding"`
+	Content         []byte `json:"-"`
 	// Crc32c checksum of Content. calculated by server when it's upload methods are used.
 	Crc32c  string            `json:"crc32c,omitempty"`
 	Md5Hash string            `json:"md5hash,omitempty"`
-	ACL     storage.ACLEntity `json:"acl,omitempty"`
+	ACL     []storage.ACLRule `json:"acl,omitempty"`
```
fsouza: Since fake-gcs-server is also used as a library, could we avoid a breaking change here? Or would the previous ``ACL`` attribute not make any sense? 🤔

lelikg: the previous ACL doesn't make sense anymore, that's why I replaced it.

lelikg: it's more compatible with GCP protocol

henrahmagix:
@lelikg I was having issues using fake-gcs-server that this pull-request solves, thanks so much for it! This fixes #137 for me.

@fsouza I'd love to have this merged so I can use it =) Maybe it's possible to split the bugfix that fixes #137 for me into a separate pull-request, and then support the predefined ACL in a separate one that can discuss the breaking change?

Thankfully I'm able to use this in my own code already, so I'm not blocked until this is merged:
```go.mod
replace github.com/fsouza/fake-gcs-server => github.com/wix-playground/fake-gcs-server v1.12.2-0.20191111164510-94f47cfa0a5d
```

fakestorage/object.go
fsouza: Yeah I was worrying about the fact that people's code will fail to compile after this change, so I wonder if we should introduce q new attribute for that.

Let me take another look at the code in a bit and check what it would take.

fakestorage/object.go
```diff
@@ -113,14 +116,17 @@ func toBackendObjects(objects []Object) []backend.Object {
 func fromBackendObjects(objects []backend.Object) []Object {
 	backendObjects := []Object{}
 	for _, o := range objects {
+		var acl []storage.ACLRule
+		_ = json.Unmarshal(o.ACL, &acl)
```
fsouza: 🤔 not sure if ignoring this error is a very good idea. If o.ACL is not empty, it should be a valid acl.

Alternatively, we can let the backend handle this and make sure that backend.Object also has ``[]storage.ACLRule``. Thoughts?

lelikg: we could return an error here but it would break the pattern. Can you show me what you mean by letting the backend handle it?

lelikg:
> @lelikg I was having issues using fake-gcs-server that this pull-request solves, thanks so much for it! This fixes #137 for me.
> 
> @fsouza I'd love to have this merged so I can use it =) Maybe it's possible to split the bugfix that fixes #137 for me into a separate pull-request, and then support the predefined ACL in a separate one that can discuss the breaking change?
> 
> Thankfully I'm able to use this in my own code already, so I'm not blocked until this is merged:
> 
> ```ampl
> replace github.com/fsouza/fake-gcs-server => github.com/wix-playground/fake-gcs-server v1.12.2-0.20191111164510-94f47cfa0a5d
> ```

glad to contribute to an already very useful piece of software

fakestorage/object.go
fsouza: I thought about making it ``[]storage.ACLRule`` in backend.Object, then implementing {Marshal,Unmarshal}JSON to properly convert it to a byte slice before marshaling. Not sure if that would work though.

If you prefer, I can merge this as is and give it a shot. Let me know what you think.

lelikg:
imho you can merge as-is :)

fsouza:
@lelikg thanks for contributing!! 😁

PR #134, opened 2019-11-18
GetObject can download content

chrisrecalis:
Getting an object through the storage api can download the content of file when query param `alt=media`. Docs: https://cloud.google.com/storage/docs/json_api/v1/objects/get


chrisrecalis:
This fixes the nodejs client as it uses the storage api to download the data

PR #140, opened 2019-11-22
Support to queries against all versions (in versioned and not versioned a buckets)

dcaba:
_note: incremental PR: the review and potential merge should go after https://github.com/fsouza/fake-gcs-server/pull/60_ 

Support for querying all versions, that means multiple generations of each file are expected in versioned buckets where there has been overwrites or deletions

Done:
* some queries with "versioning equal to true" at fakestorage level, testing a couple of scenarios
* Incorporating two simple calls to list query (one with versioning equal to false, as we had, and another with true) in the backend tests (`TestObjectCRUD`)
* implementation (easiest part actually)

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -74,8 +74,8 @@ func (s *Server) createObject(obj Object) error {
 
 // ListObjects returns a sorted list of objects that match the given criteria,
 // or an error if the bucket doesn't exist.
-func (s *Server) ListObjects(bucketName, prefix, delimiter string) ([]Object, []string, error) {
-	backendObjects, err := s.backend.ListObjects(bucketName)
+func (s *Server) ListObjects(bucketName, prefix, delimiter string, versions bool) ([]Object, []string, error) {
+	backendObjects, err := s.backend.ListObjects(bucketName, versions)
 	if err != nil {
 		return nil, nil, err
 	}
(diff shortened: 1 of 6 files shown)


dcaba:
rebased & conflicts solved, @fsouza . Thanks for the patience,

fakestorage/object.go
```diff
@@ -74,8 +74,8 @@ func (s *Server) createObject(obj Object) error {
 
 // ListObjects returns a sorted list of objects that match the given criteria,
 // or an error if the bucket doesn't exist.
-func (s *Server) ListObjects(bucketName, prefix, delimiter string) ([]Object, []string, error) {
-	backendObjects, err := s.backend.ListObjects(bucketName)
+func (s *Server) ListObjects(bucketName, prefix, delimiter string, versions bool) ([]Object, []string, error) {
```
fsouza: Should we introduce a ``ListObjectsWithOpts`` and deprecate this method?

dcaba: Hi!! Francisco, completely up to you (and I can change if you think is convenient), but I think there is an important difference between both functions: `CreateBucket` / `CreateBucketWithOpts` are probably used directly by users of this project: they are actually not published / mapped in the router of the server (but `createBucketByPost`, that calls directly createBucket in the backend - something maybe we want to change), and I guess most of the people (including me) are using that to generate scenarios/prepare a bucket status to run tests against (as the code you want to test normally expects the bucket to be already available as part of the infra/deployment setup). However, I guess `ListObjects` is mostly consumed via the API we expose in the server, so through a gcs client, and its rare people invokes it directly to do assertions... that means changes in this contract are probably not as severe as the CreateBucket, that I bet is called directly from test code. 

Do you have the same view than me? Best!!!!

fsouza: That's fair enough. Let's merge as is. Thanks for the quick response! 😁

PR #143, opened 2019-11-28
Rewrite with generation support

dcaba:
_note: again incremental PR: the review and potential merge should go after https://github.com/fsouza/fake-gcs-server/pull/140_ 

An easy one: just looking and considering the optional query arg that may come with rewrite (copy) requests to specify an specific generation of an object. Tests decoupled from the ones that work without versioning neither specific generations, so we initialize the buckets accordingly. Extra test scenario considered for rewrite object (against the same active object)

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -298,7 +298,23 @@ func (s *Server) setObjectACL(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) rewriteObject(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
-	obj, err := s.GetObject(vars["sourceBucket"], vars["sourceObject"])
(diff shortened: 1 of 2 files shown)


fakestorage/object.go
```diff
@@ -299,7 +299,23 @@ func (s *Server) setObjectACL(w http.ResponseWriter, r *http.Request) {
...
+	generationStr := r.FormValue("sourceGeneration")
+	var (
+		obj        Object
+		err        error
+		generation int64
+	)
+	if generationStr != "" {
+		generation, err = strconv.ParseInt(generationStr, 10, 64)
+		if err != nil {
+			fmt.Println(err)
```
fsouza: dd

dcaba: :+1:

dcaba:
Rebased @fsouza : it's yours :)

PR #144, opened 2019-11-29
Support for bucket creation time

dcaba:
_Again, incremental PR: to be considered after: https://github.com/fsouza/fake-gcs-server/pull/143_

This time not connected to versioning: extending the bucket representation in the memory engine to store the original bucket creation date, while in FS we are using the bucket root dir ctime to fill that field (and that is wrong, but persisting bucket metadata in that engine in a magic file looks overkilling in the current state), and extending then responses.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -70,7 +70,12 @@ func (s *Server) createBucketByPost(w http.ResponseWriter, r *http.Request) {
 	}
 
 	// Return the created bucket:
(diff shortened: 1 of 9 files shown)


dcaba:
@fsouza , this has been rebased onto https://github.com/fsouza/fake-gcs-server/pull/143 . If you merge without squashing that one, this should be ready (up to you! can rebase again if needed)

fsouza:
Thanks! Will cut a new release after merging.

PR #155, opened 2019-12-28
fakestorage: add logging to server

cuonglm:
It helps debugging requests to server easier.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -8,6 +8,7 @@ import (
 	"context"
 	"crypto/tls"
 	"fmt"
+	"io"
 	"net"
 	"net/http"
 	"net/http/httptest"
@@ -15,6 +16,7 @@ import (
 
 	"cloud.google.com/go/storage"
 	"github.com/fsouza/fake-gcs-server/internal/backend"
+	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
 	"google.golang.org/api/option"
 )
@@ -77,6 +79,9 @@ type Options struct {
 	// https://<bucket>.storage.gcs.127.0.0.1.nip.io:4443>/<bucket>/<object>
 	// If unset, the default is "storage.googleapis.com", the XML API
 	PublicHost string
+
+	// Destination for writing log.
+	Writer io.Writer
 }
 
 // NewServerWithOptions creates a new server with custom options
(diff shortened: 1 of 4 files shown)


cuonglm:
Hey @fsouza, a gentle ping 😄

fakestorage/server.go
```diff
@@ -14,9 +14,12 @@ import (
 	"sync"
 
 	"cloud.google.com/go/storage"
-	"github.com/fsouza/fake-gcs-server/internal/backend"
+	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
+	"github.com/sirupsen/logrus"
 	"google.golang.org/api/option"
+
+	"github.com/fsouza/fake-gcs-server/internal/backend"
```
fsouza: Let's keep just two groups 😁

fakestorage/server.go
```diff
@@ -90,7 +93,7 @@ func NewServerWithOptions(options Options) (*Server, error) {
 		return s, nil
 	}
 
-	s.ts = httptest.NewUnstartedServer(s.mux)
+	s.ts = httptest.NewUnstartedServer(handlers.LoggingHandler(logrus.New().Writer(), s.mux))
```
fsouza: Can you introduce a knob for this and make it off by default? Just want to avoid polluting stderr for projects using fake-gcs-server as a library.

We can add an ``io.Writer`` to ``Options`` for logging  and whenever that's not nil, we setup the LoggingHandler.

Then in the main package, we can pass ``logrus.New().Writer()`` as the writer. That way, users of the library won't get any logs unless they explicitly ask for it, and users of the daemon will get logs. What do you think?

cuonglm: 👍

cuonglm: @fsouza I think a bool flag is enough

fakestorage/server.go
cuonglm: FIxed

fakestorage/server.go
fsouza: @cuonglm I can see people asking for the ability to customize where the logs go to (stdout vs stderr, for example) and if we decide to implement that, we'd need yet another parameter, but passing an writer makes that use case possible and avoids the direct dependency on logrus from the library package (along with allowing library users to pass an arbitrary writer if they need an in-memory buffer or something like that for some reason 😁).

cuonglm: @fsouza Ah ok, just because I see we're using logrus already. My concern is that we can't pass the io.Writer via command line argument.

cuonglm: @fsouza fixed

PR #189, opened 2020-03-03
Implement minimal support for PATCH metadata API

snawaz:
This PR implements minimal support for `PATCH` object's metadata API.

It patches the object's metadata as per `path parameters` and `metadata` in the request body. It  _completely_ ignores the `optional query parameters` of the API, which are slightly complex. So if this minimal support is acceptable, then we can merge it, and the optional query parameters can be taken care of in later PRs (if needed).

API Ref: https://cloud.google.com/storage/docs/json_api/v1/objects/patch

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -399,3 +399,25 @@ func (s *Server) handleRange(obj Object, r *http.Request) (start, end int, conte
 	}
 	return 0, 0, obj.Content
 }
+
+func (s *Server) patchObject(w http.ResponseWriter, r *http.Request) {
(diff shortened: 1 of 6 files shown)


snawaz:
> This looks good! Can you add a test that uses [ObjectHandle.Update](https://pkg.go.dev/cloud.google.com/go/storage?tab=doc#ObjectHandle.Update) just to confirm it works and prevent regressions?

Sure.

snawaz:
@fsouza 

The test has been added. Please have a look at the second commit : `Add test for PATCH API`.

PR #229, opened 2020-04-30
github/workflows/main: add windows to CI

fsouza:


kennytm:
Just wanna note that the Windows version of git by default won't create symlinks (they are created as regular files with the target as content), which is why `TestGenerateObjectsFromFiles/should_support_multiple_levels` is failing.

PR #240, opened 2020-05-08
fakestorage: misc fixes

### LINKED ISSUE #239: Expected headers are not sent for ObjectHandle.NewReader()
The [Reader](https://godoc.org/cloud.google.com/go/storage#Reader) you get from `ObjectHandle.NewReader` has an `Attrs` field of type [ReaderObjectAttrs](https://godoc.org/cloud.google.com/go/storage#ReaderObjectAttrs).

The `storage` library fills those values in based on headers sent in the response: 

https://github.com/googleapis/google-cloud-go/blob/bd00477b309def1610fbf4cb259329deac795891/storage/reader.go#L255-L268

As far as I can tell, `fakestorage` is not setting these headers, which means that you get something like the following when you try to use them:

```
reader attrs: {Size:56 StartOffset:0 ContentType: ContentEncoding: CacheControl: LastModified:0001-01-01 00:00:00 +0000 UTC Generation:-1 Metageneration:0}
```

This library has been great, but including these headers would make it so that I can safely use the `Attrs` attribute _and_ be able to test. Thanks!

someone1:
* Add Last-Modified and X-Goog-Generation HTTP headers for GET/HEAD object requests
* Fallback to original object's metadata on rewrite requests
* Return object generation number if available after upload
* Fix Object Generation JSON encoding to align with Go library

Signed-off-by: Prateek Malhotra <someone1@gmail.com>

Fixes #239

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -65,13 +65,11 @@ func (o *objectList) Swap(i int, j int) {
 //
 // If the bucket within the object doesn't exist, it also creates it. If the
 // object already exists, it overrides the object.
-func (s *Server) CreateObject(obj Object) Object {
-	newObj, err := s.createObject(obj)
+func (s *Server) CreateObject(obj Object) {
+	_, err := s.createObject(obj)
 	if err != nil {
 		panic(err)
 	}
-
-	return newObj
 }
 
 func (s *Server) createObject(obj Object) (Object, error) {
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -372,7 +372,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Accept-Ranges", "bytes")
 	w.Header().Set("Content-Length", strconv.Itoa(len(content)))
 	w.Header().Set(contentTypeHeader, obj.ContentType)
-	w.Header().Set("X-Goog-Generation", fmt.Sprintf("%d", obj.Generation))
+	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
 	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
 	w.WriteHeader(status)
 	if r.Method == http.MethodGet {
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -65,15 +65,22 @@ func (o *objectList) Swap(i int, j int) {
 //
 // If the bucket within the object doesn't exist, it also creates it. If the
 // object already exists, it overrides the object.
-func (s *Server) CreateObject(obj Object) {
-	err := s.createObject(obj)
+func (s *Server) CreateObject(obj Object) Object {
+	newObj, err := s.createObject(obj)
 	if err != nil {
 		panic(err)
 	}
+
+	return newObj
 }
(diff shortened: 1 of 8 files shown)


fakestorage/upload.go
```diff
@@ -197,6 +197,11 @@ func (s *Server) multipartUpload(bucketName string, w http.ResponseWriter, r *ht
 		http.Error(w, err.Error(), http.StatusInternalServerError)
 		return
 	}
+
+	// If the backend created a Generation #, add it to the response
+	newObj, _ := s.GetObject(obj.BucketName, obj.Name)
+	obj.Generation = newObj.Generation
```
fsouza, line 203: I think it makes more sense to modify createObject to return the created object. What do you think?

fakestorage/object.go
```diff
@@ -361,6 +372,8 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Accept-Ranges", "bytes")
 	w.Header().Set("Content-Length", strconv.Itoa(len(content)))
 	w.Header().Set(contentTypeHeader, obj.ContentType)
+	w.Header().Set("X-Goog-Generation", fmt.Sprintf("%d", obj.Generation))
```
fsouza, line 375: Let's use strconv instead:

```suggestion
	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
```

fakestorage/upload.go
someone1: Agreed - I didn't want to introduce any larger changes though. I'll update accordingly.

Speaking of larger changes, is there a reason you didn't reuse the [Object](https://godoc.org/google.golang.org/api/storage/v1#Object) type provided by the Google API code instead of re-creating your own `Object` type? I'd be willing to tackle this conversion as well.

fsouza: I don't remember exactly why, maybe the size of the struct or something like that. Let's keep the custom Object type as changing that would be an API breaking change.

someone1: Ok - I added the changes for having `createObject` return the created object. 

There should be virtually 100% overlap between the Object types - I think some data captured in `fake-gcs-server` is different/omitted in the JSON as compared to the original `Object` (e.g. `Bucket` vs `BucketName`). Might be worth considering, would have also caught the encoding issue with the `Generation` attribute.

someone1:
I actually have one more change incoming, to (barely) support signed uploads

someone1:
@fsouza - should be ready for review now!

fakestorage/object.go
```diff
@@ -65,15 +65,22 @@ func (o *objectList) Swap(i int, j int) {
 //
 // If the bucket within the object doesn't exist, it also creates it. If the
 // object already exists, it overrides the object.
-func (s *Server) CreateObject(obj Object) {
-	err := s.createObject(obj)
+func (s *Server) CreateObject(obj Object) Object {
```
fsouza, line 68: Let's not change this once because it's part of the public API. We should probably deprecate this in favor of one that returns the object and an error instead of panic, but it doesn't need to be part of this PR.

someone1: Reverted.

fsouza:
Thanks!

PR #243, opened 2020-05-12
Add EXPOSE to Dockerfile

simondrake:
## What is being changed

Add an `EXPOSE` command to the projects `Dockerfile`

## Why are we making this change?

In order to use the Docker image in CI pipelines, it's imperative that the `EXPOSE` command is in the `Dockerfile`. 

For example, see [How the health check of services works](https://docs.gitlab.com/ee/ci/docker/using_docker_images.html#how-the-health-check-of-services-works) from the Gitlab CI runner documentation. Specifically, the part which states:

> Checks which ports are exposed from the container by default

Without this, the pipeline will fail with an invalid host/port error. 

This change has also been published to [Docker Hub](https://hub.docker.com/repository/docker/omisnomis/fake-gcs-server) to prove that it now works as expected.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,4 @@
 fake-gcs-server
+
+# editor and IDE
+.idea
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,1 @@
 fake-gcs-server
-
-# editor and IDE
-.idea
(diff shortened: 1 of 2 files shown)


.gitignore
```diff
@@ -1 +1,4 @@
 fake-gcs-server
+
+# editor and IDE
+.idea
```
fsouza, line 4: Can you remove this? For editor specific files, I recommend something like a global ignore file.

simondrake: Sure. Done.

PR #266, opened 2020-06-08
Add check to ensure Range is less than content

gjabell:
* This fixes the case where a Range value greater than the length of the
  content is requested, causing the server to crash and fail the
  request.


gjabell:
Hey @fsouza, can do :)

gjabell:
Let me know how those look @fsouza; the `Too-long range specified` test should fail without the fix and pass with it :)

PR #286, opened 2020-07-12
Allow to pass custom public host to docker image

borovik96:


borovik96:
How can I start docker image https://hub.docker.com/r/fsouza/fake-gcs-server with custom `publicHost` and `http` schema?

fsouza:
@borovik96 you can do `docker run fsouza/fake-gcs-server -public-host <something> -scheme http`

borovik96:
Yes, you are right! Thanks for help

PR #306, opened 2020-08-12
Update README.md with possible keywords

antonydenyer:
I struggled to find this as I was looking for a stub rather than an emulator.


fsouza:
@antonydenyer thank you very much for contributing!

PR #307, opened 2020-08-12
add support for MD5 checksum of object content when using fs

antonydenyer:
Copy pasted md5 and crc hash functions from `upload.go`.

I compared the results against actually uploading to GCS (and using `gsutil hash -m`) and they differ for some reason. Whilst this is somewhat annoying it isn't a problem for me as all I need is something consistent.

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -5,9 +5,12 @@
 package backend
 
 import (
+	"crypto/md5" // #nosec G501
+	"encoding/base64"
 	"encoding/json"
 	"errors"
 	"fmt"
+	"hash/crc32"
 	"io/ioutil"
 	"net/url"
 	"os"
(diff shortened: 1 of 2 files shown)


internal/backend/fs.go
```diff
@@ -163,6 +166,37 @@ func (s *storageFS) GetObjectWithGeneration(bucketName, objectName string, gener
...
 
+var crc32cTable = crc32.MakeTable(crc32.Castagnoli)
+
+func crc32cChecksum(content []byte) []byte {
+	checksummer := crc32.New(crc32cTable)
+	checksummer.Write(content)
+	return checksummer.Sum(make([]byte, 0, 4))
+}
+
+func encodedChecksum(checksum []byte) string {
```
fsouza, line 177: Since this and `encodedHash` are the same, could we unify them (with a common name)? Or maybe inline the call to `base64.StdEncoding.EncodeToString`?

internal/backend/fs.go
```diff
@@ -175,6 +209,8 @@ func (s *storageFS) getObject(bucketName, objectName string) (Object, error) {
 	}
 	obj.Name = filepath.ToSlash(objectName)
 	obj.BucketName = bucketName
+	obj.Crc32c = encodedCrc32cChecksum(obj.Content)
+	obj.Md5Hash = encodedMd5Hash(obj.Content)
```
fsouza, line 213: I wonder if we should generate those on creation and store in the JSON. What do you think?

antonydenyer: Disagree, if you start the server then overwrite a file you would have a hash mismatch. You could lock the files but I think it's overkill.

internal/backend/fs.go
antonydenyer: Can do, like I said it's a copy and paste job. Ideally, the md5 and crc code should be extracted and used by `fs.go` and `upload.go` but my golang knowledge is poor!

internal/backend/fs.go
fsouza: One of the points of a checksum is to ensure integrity, and an external process overwriting a file is a break of integrity x) It's a test server though, so no big deal.

fsouza: Oh I merged, but I'm going to revert. The server already behaves accordingly as I suggested, I don't think it's a good idea to have storage backends behave differently on data.

I created an issue for generating the md5 and crc sums when seeding from the disk (see #313).

PR #318, opened 2020-08-28
Support CNAME'd buckets

kenniaa:
I was having some issues interfacing with `fake-gcs-server` for buckets behind a CNAME, where the bucket name is the host itself. It seems as though only Path based (`storage.googleapis.com/myBucketName`) or VHost based (`myBucketName.storage.googleapis.com`) are supported.

This small fix addresses the issue by providing additional routes which consider the host as a bucket name; one for GETing a file and another for Signed URLs.

I also had to shuffle the route order slightly to prevent routes from conflicting.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -184,12 +184,14 @@ func (s *Server) buildMuxer() {
 		r.Path("/b/{sourceBucket}/o/{sourceObject:.+}/rewriteTo/b/{destinationBucket}/o/{destinationObject:.+}").HandlerFunc(s.rewriteObject)
 	}
 
-	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
 	bucketHost := fmt.Sprintf("{bucketName}.%s", s.publicHost)
 	s.mux.Host(bucketHost).Path("/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
 	s.mux.Path("/download/storage/v1/b/{bucketName}/o/{objectName:.+}").Methods("GET").HandlerFunc(s.downloadObject)
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods("POST").HandlerFunc(s.insertObject)
 	s.mux.Path("/upload/resumable/{uploadId}").Methods("PUT", "POST").HandlerFunc(s.uploadFileContent)
+
+	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
+	s.mux.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
 
 	// Signed URL Uploads
 	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -194,6 +194,7 @@ func (s *Server) buildMuxer() {
 	// Signed URL Uploads
 	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
 	s.mux.Host(bucketHost).Path("/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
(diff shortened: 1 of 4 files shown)


fsouza:
Hi @kenniaa, I'm not 100% familiar with the use case. Can you add a test or provide an example of how you use this, just to make sure we get the implementation right?

kenniaa:
Hey!

I added two tests to cover the added behavior-- sorry for not doing so before.

To clarify the intent behind the PR, with GCS you can name a bucket something like `mybucket.mydomain.com` which allows you to interact with the bucket via a CNAME to `c.storage.googleapis.com`. Here is the [official documentation][0] regarding this to explain further.

The current behavior in `fake-gcs-server` gets us _really_ close to mocking this specific use case, but falls just short of doing it. With the changes in this PR, you can now make a request to `fake-gcs-server` where the bucket name is the host and file gets and (signed) uploads will work, effectively mimicking the actual prod behavior.

[0]: https://cloud.google.com/storage/docs/request-endpoints#cname

fakestorage/server.go
```diff
@@ -184,16 +184,19 @@ func (s *Server) buildMuxer() {
...
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods("POST").HandlerFunc(s.insertObject)
 	s.mux.Path("/upload/resumable/{uploadId}").Methods("PUT", "POST").HandlerFunc(s.uploadFileContent)
 
+	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
+	s.mux.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
+
 	// Signed URL Uploads
 	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
 	s.mux.Host(bucketHost).Path("/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
+	s.mux.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods("POST", "PUT").HandlerFunc(s.insertObject)
```
fsouza, line 199: You can upload using the custom cname too? Does that mean GCS will accept uploads over http in plain-text mode? 😱

fsouza (approved):
Thank you for contributing and sorry for the delayed review.

This looks good overall, just a couple of inline question plus this one: should we be strict about this and make it work only over HTTP (vs HTTPS), like GCS does? Or should we not care?

fakestorage/server.go
kenniaa: Yup! It's surprising because it's not well documented, but it actually works.

And yeah, unfortunately :(

In production for me, I get "around" this by doing SSL termination at the CDN but ultimately the traffic to the bucket is HTTP not HTTPS.

fakestorage/server.go
```diff
@@ -184,16 +184,19 @@ func (s *Server) buildMuxer() {
...
 
-	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
 	bucketHost := fmt.Sprintf("{bucketName}.%s", s.publicHost)
 	s.mux.Host(bucketHost).Path("/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
 	s.mux.Path("/download/storage/v1/b/{bucketName}/o/{objectName:.+}").Methods("GET").HandlerFunc(s.downloadObject)
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods("POST").HandlerFunc(s.insertObject)
 	s.mux.Path("/upload/resumable/{uploadId}").Methods("PUT", "POST").HandlerFunc(s.uploadFileContent)
 
+	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
+	s.mux.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods("GET", "HEAD").HandlerFunc(s.downloadObject)
```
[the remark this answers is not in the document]
kenniaa: At first, I agreed with you 100% on this since a CNAME would only allow for `<subdomain>.<domain>.<tld>`, but then I realized two key things:

- GCS supports [A records](https://cloud.google.com/storage/docs/naming-buckets) as well (with A redirects)
- CNAME flattening exists 

So I *think* it might be possible to have any hostname be valid in this context. I'm not 100% sure here, so I'll leave it up to you to decide which direction we should go in.

kenniaa:
> Should we be strict about this and make it work only over HTTP (vs HTTPS), like GCS does? Or should we not care?

I think we should err on the side of not caring-- since `fake-gcs-server` is going to be used only in dev or qa contexts, it might be best to leave it as open as possible.

PR #319, opened 2020-08-30
Add CORS Support

kenniaa:
A big reason for supporting signed urls is the ability to initiate requests from the browser that insert files into the bucket instead of proxying files through your server.

Currently, this does not work in `fake-gcs-server` due to CORS limitations, as `fake-gcs-server` does not respond to any CORS preflight requests. From the browser, you may disable CORS checking (via `no-cors` on `fetch` or equivalent), but this [severely limits][no-cors-limits] what kind of requests you can make; with CORS checking disabled, you are unable to initiate a PUT request thereby preventing the above usecase.

So to address this, I've added a simple CORS handler. It is intended to be maximally accepting, such that most requests will be accepted without configuration. Since this project is meant for testing GCS and not running in production, I did not see much of a reason for locking it down and instead opted for ease of use. 

The one exception to this is `Access-Control-Allowed-Headers`, because unfortunately Mux [will not support][header-wildcard-issue] wildcards for headers. This means that I instead had to audit what headers `fake-gcs-server` accepts and manually specify them in the allowlist-- I believe I have covered them all but I could be missing something (CORS requests never worked here, so I don't consider that a dealbreaker). Additionally, I've added a new command line flag for the end user to specify any additional headers they need to be present in the allowlist. I found this useful for the custom meta headers that you may attach to signed urls.

I've added tests which should cover the necessary changes to config and server function, but please let me know if I should add any others!

[no-cors-limits]: https://developer.mozilla.org/en-US/docs/Web/API/Request/mode
[header-wildcard-issue]: https://github.com/gorilla/handlers/pull/180


internal/config/config.go
```diff
@@ -43,11 +47,20 @@ func Load(args []string) (Config, error) {
...
 	err := fs.Parse(args)
 	if err != nil {
 		return cfg, err
 	}
+
+	if len(allowedCORSHeaders) > 0 {
+		cfg.allowedCORSHeaders = strings.Split(allowedCORSHeaders, ",")
+	} else {
+		cfg.allowedCORSHeaders = []string{}
+	}
```
fsouza, line 62: Let's check if the string is not empty and remove the else block.

```suggestion
	if allowedCORSHeaders != "" {
		cfg.allowedCORSHeaders = strings.Split(allowedCORSHeaders, ",")
	}
```

internal/config/config_test.go
```diff
@@ -27,33 +27,36 @@ func TestLoadConfig(t *testing.T) {
...
-				externalURL: "",
-				host:        "0.0.0.0",
-				port:        4443,
-				scheme:      "https",
+				Seed:               "",
+				backend:            "filesystem",
+				fsRoot:             "/storage",
+				publicHost:         "storage.googleapis.com",
+				externalURL:        "",
+				allowedCORSHeaders: []string{},
```
fsouza, line 56:
```suggestion
				allowedCORSHeaders: nil,
```

fsouza (approved):
Looks good overall, just one nitpicky comment.

Too bad gorilla won't take `*` for allowed-headers

internal/config/config.go
kenniaa: Ha! What was I thinking 🤦🏽

kenniaa:
Interesting that this only failed the lint on Windows 🤔

fsouza:
The linting thing was a lot of work to debug, will open some GH issues later today to try to help project authors improve the error reporting (when using the gh action, if linting fails on a file that's not modified in the PR, there's no way to detect where the error is coming from) and the false positive (the reported error isn't really an issue).

Anyways, can you rebase your branch off main? That should fix the build, then we can ship this!

fsouza (approved):
Actually let me just merge this and it will pass when merged.

PR #330, opened 2020-09-17
Fix bucket name validation to allow underscores

gh2k:
According to Google's [bucket naming guidelines](https://cloud.google.com/storage/docs/naming-buckets) underscores in bucket names are OK.

This patch updates the validation regex to more closely match the guidelines, by:

* Allowing underscores
* Preventing non-alphanumeric characters at the start/end of names

I haven't added the length checks, or for the `goog` prefix, so these can still be created erroneously - but this should fix downstream projects that have broken because of the incorrect name validation change.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -13,7 +13,7 @@ import (
 	"github.com/gorilla/mux"
 )
 
-var bucketRegexp = regexp.MustCompile(`^[a-zA-Z0-9-][a-zA-Z0-9.-]*[a-zA-Z0-9-]$`)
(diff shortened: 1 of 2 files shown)


gh2k:
You can have [underscores in DNS](https://stackoverflow.com/questions/2180465/can-domain-name-subdomains-have-an-underscore-in-it#2183140), you just can't register a domain with one in from most (all?) registrars. So, they work fine in subdomains. :-)

fsouza:
Oh sorry I mean the hostname. I guess Google's reverse proxy isn't strict about hostnames 🤷

PR #345, opened 2020-10-19
Improved nodejs example

jeantil:
* This example works as is using an identity associated to a valid billing account
* This example works using a free firebase account (which provides a free bucket) but you need to comment the `createBucket` and `bucket.delete()` since free accounts are not allowed to create additional buckets
* This example works against fake-gcs-server if you configure free firebase account credentials when creating the storage instance.

Note that the following fails : 

- without any credentials (as is suggested in the original example) file manipulation operations fail with : 
```
Error: Could not load the default credentials. Browse to https://cloud.google.com/docs/authentication/getting-started for more information.
    at GoogleAuth.getApplicationDefaultAsync (/home/jean/dev/startups/yupwego/src/fake-gcs-server/examples/node/node_modules/google-auth-library/build/src/auth/googleauth.js:155:19)
```
this has been reported and closed before as being a client library error, I have reported it at https://github.com/googleapis/nodejs-storage/issues/1315

### DIFF
--- a/examples/node/index.js
+++ b/examples/node/index.js
@@ -1,26 +1,113 @@
 process.env["NODE_TLS_REJECT_UNAUTHORIZED"] = 0;
+const globalLog = require("global-request-logger");
+const chalk = require("chalk");
+const DEBUG_HTTP = 0;
+globalLog.initialize();
 
-async function listBuckets() {
-  // [START storage_list_buckets]
+function httpDebug(msg, ...args) {
+  if (DEBUG_HTTP) console.log(chalk.blue(msg), ...args);
+}
+
+function logHeaders(headers) {
+  if (headers) {
+    Object.entries(headers).forEach(([key, value]) =>
+      typeof value.forEach === "function"
+        ? value.forEach((v) =>
+            httpDebug(
+              "%s: %s",
+              key,
+              key.toLowerCase() === "authorization" ? "[...redacted...]" : v
+            )
+          )
+        : httpDebug("%s: %s", key, value)
+    );
+  }
+}
+function logRequestAndResponse(request, response) {
+  httpDebug("");
+  httpDebug("%s %s", request.method, request.path);
+  logHeaders(request.headers);
+  if (request.body) httpDebug(request.body);
+  if (response && response.httpVersion) {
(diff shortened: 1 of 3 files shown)


jeantil:
the upstream team has proposed a fix for the customEndpoint issue in https://github.com/googleapis/nodejs-storage/pull/1316 once merged I'll upgrade the dependencies in this PR and the checks should pass

fsouza:
Hi @jeantil, do you still plan to work on this? https://github.com/googleapis/nodejs-storage/pull/1316 is merged and released, so we should already have it :D Let me know how you'd like to proceed.

jeantil:
@fsouza I completely missed your comment at the time. 
if you updated the dependency to use the fixed nodejs-storage you might want to use the example in this PR instead of the one currently in https://github.com/fsouza/fake-gcs-server/blob/main/examples/node/index.js 
The example in this PR is more complete and demonstrates a broader usage of the fake-gcs-server

fsouza:
@jeantil no problem, I'll reopen it and pick it back some time this week or next week.

fsouza:
Hey @jeantil, I'm a bit confused by what's going on here. What exactly requires the Firebase free account? The current example downloads the file which I think is enough? Sorry I probably haven't touched that example in a while 🙈 

I don't think examples need to test the whole thing 🤔 I'm going to close this for now, but let me know if you'd want to move forward with it by fixing the merge conflict and then I can take another look.

jeantil:
> The current example downloads the file which I think is enough[...]I don't think examples need to test the whole thing.
It's your project :) I personally prefer to have as full a contract test as possible. The example currently included in the codebase only does read operations and I felt it is too little.

> What exactly requires the Firebase free account?
The benefit of a contract test it to make it easy to compare the behaviour of different implementations (fake-gcs vs actual gcs for instance) 
Firebase free accounts include a free GCS bucket which is a nice way to experiment with GCS for free (IIRC you don't even need to provide credit card information).
However the GCS bucket comes predefined and you cannot create a new one, hence the comment. 
this part can definitely be dropped, I included it so that you/other contributors could easily compare the reported issue with the actual behavior.

Issue #222, opened 2020-04-22
Does that support storage events notifications ? 


Ilchuk-Mihail:
Trying to implement [pubsub-notifications](https://cloud.google.com/storage/docs/pubsub-notifications) and receive notifications when new file is stored in bucket.

fdescamps:
Hi,

Did you succeed to listen storage triggers event (like an upload on a bucket) please ?

Thanks

Ilchuk-Mihail:
@fdescamps It was not possible there, so I ended up using my custom solution. (lib which stores files in tmp folder and fire pubsub event)

fdescamps:
Hi,

Thanks for your quick answer.
I'm pretty "sad" about this unavailable feature :/
If anyone manages this, I would like to have some tips.

Regards,

Ilchuk-Mihail:
I see there opened pull request https://github.com/fsouza/fake-gcs-server/pull/370 with this. Hope they will merge it soon.

mjeffrey:
I think this can be closed. I am using it without an issue.

PR #370, opened 2020-11-26
Added support to send events to a pubsub topic

isaldana:
This addresses #222.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -17,20 +17,27 @@ import (
 )
 
 const (
-	filesystemBackend = "filesystem"
-	memoryBackend     = "memory"
+	filesystemBackend   = "filesystem"
+	memoryBackend       = "memory"
+	eventFinalize       = "finalize"
+	eventDelete         = "delete"
+	eventMetadataUpdate = "metadataUpdate"
 )
 
 type Config struct {
-	Seed               string
-	publicHost         string
-	externalURL        string
-	allowedCORSHeaders []string
-	scheme             string
-	host               string
-	port               uint
-	backend            string
-	fsRoot             string
+	Seed                string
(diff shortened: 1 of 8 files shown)


isaldana:
It looks like the lint test is failing because the the `Dockerfile` needs to be updated to freeze ` golangci/golangci-lint:v1.32 AS linter` since the newest version breaks the existing code (even without my PR). I updated it in this PR.

fsouza:
Hi  @isaldana thanks for contributing and apologies for such a delay to get back to you. I'll look into this PR later today or tomorrow!

internal/config/config.go
```diff
@@ -17,27 +17,35 @@ import (
...
+	allowedCORSHeaders  []string
+	scheme              string
+	host                string
+	port                uint
+	backend             string
+	fsRoot              string
+	eventPubsubProjecID string
+	eventPubsubTopic    string
+	eventPrefix         string
+	eventList           []string
```
fsouza, line 40: Can you move these fields into a dedicated struct for events?

fsouza: (also, with the growing number of flags we have, we'd should probably consider adding some form of config file, not a blocker for this PR though, definitely something I'll have to figure out later)

fsouza (commented):
Sorry for such a terrible turnaround on this. I've rebased it and gave it a first pass. Looks good overall, still thinking about the idea though. It'd be good if this could be 100% mocked or something like that (perhaps we can leverage the eventPublisher interface for more than just tests). Since this is all internal state, we don't need to block the PR on that decision.

Also, if you've moved on to work on other stuff, let me know and I pickup the changes to get this merged!

fsouza:
Closing in favor of #490.

PR #372, opened 2020-11-28
Lock linter version to fix new failures.

mheffner:
Lock the linter to version v1.32 to prevent failures on more recent linter versions. This gets the main branch building again and would isolate future lint cleanups.

Grabbed change from #370


fsouza (approved):

fsouza:
@mheffner thanks for contributing!

Issue #371, opened 2020-11-26
Invalid Request when calling create_file with Ruby GCS SDK


mheffner:
When using the [googe-cloud-storage](https://rubygems.org/gems/google-cloud-storage/versions/0.20.0) Ruby gem, version 1.29.1, to try and upload a file to fake-gcs-server we are seeing an Invalid Request error. This is the request as seen from the fake-gcs-server logs:
```
nf-originci-gcs       | time="2020-11-26T15:17:57Z" level=info msg="172.31.0.9 - - [26/Nov/2020:15:17:57 +0000] \"POST /upload/storage/v1/b/deploy-manifests/o?name=my-key HTTP/1.1\" 400 19"
```
It would appear that we are hitting this error case:
https://github.com/fsouza/fake-gcs-server/blob/e8bd4c7f4a67f6c06a70e73be879cc981d4c84bc/fakestorage/upload.go#L71
which makes sense as we are not using a policy nor is there an `uploadType` POST param.

I can reproduce this with the following code:
```ruby
require "google/cloud/storage"

storage = Google::Cloud::Storage.anonymous(endpoint: "http://gcs:4443/")
bucket = storage.bucket("deploy-manifests", skip_lookup: true)

data = StringIO.new('{"hi":"bye"}')

r = bucket.create_file(data, "my-key", content_type: "application/json")
```

This will throw an exception in:
```
/usr/local/bundle/ruby/2.7.0/gems/google-api-client-0.48.0/lib/google/apis/core/http_command.rb:228:in `check_status': Invalid request (Google::Apis::ClientError)
	from /usr/local/bundle/ruby/2.7.0/gems/google-api-client-0.48.0/lib/google/apis/core/api_command.rb:119:in `check_status'
	from /usr/local/bundle/ruby/2.7.0/gems/google-api-client-0.48.0/lib/google/apis/core/http_command.rb:194:in `process_response'
	from /usr/local/bundle/ruby/2.7.0/gems/google-api-client-0.48.0/lib/google/apis/core/upload.rb:174:in `process_response'
	from /usr/local/bundle/ruby/2.7.0/gems/google-api-client-0.48.0/lib/google/apis/core/upload.rb:250:in `execute_once'
```

If I run a netcat locally and point the endpoint at it, this is the full initial POST request:
```
POST /upload/storage/v1/b/deploy-manifests/o?name=my-key HTTP/1.1
[...]

PR #373, opened 2020-11-30
Support undocumented resumable upload API

mheffner:
This is a simple handling of the resumable upload protocol, minimally supporting the client implementation from: https://github.com/googleapis/google-api-ruby-client/blob/master/lib/google/apis/core/upload.rb. This protocol closely follows the Photo's API protocol documented here: https://developers.google.com/photos/library/guides/resumable-uploads. This is a known undocumented API: https://github.com/googleapis/google-api-ruby-client/issues/311.

This does not implement the query method or resumable uploads from partial uploaded content. It is assumed that the entire upload is performed as a single chunk successfully.

Fixes: #371

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -25,6 +25,12 @@ import (
 
 const contentTypeHeader = "Content-Type"
 
+const (
+	uploadTypeMedia     = "media"
+	uploadTypeMultipart = "multipart"
+	uploadTypeResumable = "resumable"
+)
+
 type multipartMetadata struct {
 	ContentType     string            `json:"contentType"`
 	ContentEncoding string            `json:"contentEncoding"`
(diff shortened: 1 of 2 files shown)


fakestorage/upload.go
```diff
@@ -345,6 +359,7 @@ func (s *Server) resumableUpload(bucketName string, w http.ResponseWriter, r *ht
 // it can't process a native "308 Permanent Redirect". The in-process response
 // then has a status of "200 OK", with a header "X-Http-Status-Code-Override"
 // set to "308".
+//nolint:funlen
```
fsouza, line 362: I just pushed a change on the main branch disabling funlen globally, can you remove this line? (sorry that our golangci-lint config is so annoying, I'm planning to eventually review it and simplify it - or just go with staticcheck + some very light checks in golangci-lint - mainly because gofmt won't exit with a non-0 status when it fails :/)

fsouza (approved):
Hi @mheffner, thank you very much for contributing and apologies for such a delayed review. I'll actually merge and take care of the nolint change, then tag a new release.

fakestorage/upload.go
mheffner: Thanks @fsouza! Looks like you already got that one? https://github.com/fsouza/fake-gcs-server/commit/6ef4f624250ad255d3d435c725ce858c54b1a66c

fsouza: Yep, we're all good!

Issue #420, opened 2021-02-12
Response comply with API for object insert


tarsisazevedo:
I am using java's gcs client and it expects some data from the response that fake-gcs-server dont provide. I think api should send a response with data expected, like described in this document https://cloud.google.com/storage/docs/json_api/v1/objects#resource

If you are ok with that, I can send a PR to fix this issue.

fsouza:
Go for it. I assume we have all the data? Let me know if you run into any issues.

fsouza:
This was fixed in #843.

PR #421, opened 2021-02-14
object: fix acl and project team json marshalling

visualage:
This partially address #420. The current ACL list does not comply with the response marshalling because the JSON field names are different than the `storage.ACLRule` field names. Created a type alias `ACLRule` from `storage.ACLRule`, and added custom Marshalling. Did same for `storage.ProjectTeam`.


fsouza (commented):
I'm afraid this causes a breaking change for people using fake-gcs-server as a library :(

A safer approach would be to implement {Marshal,Unmarshal}JSON in the `Object` type.

visualage:
Done. Added custom marshaller to Object instead of changing the ACL type in Object.

fakestorage/object.go
```diff
@@ -24,24 +24,174 @@ var errInvalidGeneration = errors.New("invalid generation ID")
...
+	o.ACL = make([]storage.ACLRule, len(temp.ACL))
+	for i, ACL := range temp.ACL {
+		o.ACL[i] = storage.ACLRule(ACL)
+	}
+
+	return nil
+}
+
+// ACLRule is an alias of storage.ACLRule to have custom JSON marshal
+type ACLRule storage.ACLRule
```
fsouza, line 123: Let's not export this type since it's only used for marshaling/unmarshaling.
```suggestion
type aclRule storage.ACLRule
```

fakestorage/object.go
```diff
@@ -24,24 +24,174 @@ var errInvalidGeneration = errors.New("invalid generation ID")
...
+
+// ProjectTeam is an alias of storage.ProjectTeam to have custom JSON marshal
+type ProjectTeam storage.ProjectTeam
```
fsouza, line 126: Same as above:

```suggestion
type projectTeam storage.ProjectTeam
```

fsouza (commented):
Looks good overall, let's unexport the types that are used just for marshaling. Also, can you add some tests?

fsouza:
@tarsisazevedo does this PR fix the issue you were observing?

tarsisazevedo:
I think so @fsouza. It will send all infos in the http response as the clients expect.

fsouza (approved):
Thank you very much! I think I'll add an example that uses the Java SDK later and that should take care of testing.

PR #434, opened 2021-02-22
add put method for object updates

zdunecki:
GCS has `PUT` method to object updates. Under the hood fake-gcs-server uses `Update` method from Google but this call `PUT` request instead of `PATCH` and fake-gcs-server serves only `PATCH`, then for consistency I've added `PUT` method also.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -201,7 +201,7 @@ func (s *Server) buildMuxer() {
 		r.Path("/b/{bucketName}").Methods("GET").HandlerFunc(s.getBucket)
 		r.Path("/b/{bucketName}/o").Methods("GET").HandlerFunc(s.listObjects)
 		r.Path("/b/{bucketName}/o").Methods("POST").HandlerFunc(s.insertObject)
-		r.Path("/b/{bucketName}/o/{objectName:.+}").Methods("PATCH").HandlerFunc(s.patchObject)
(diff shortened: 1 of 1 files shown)


fsouza (commented):
Hi @zdunecki, thanks for contributing! Update and patch are not the same thing, see the note in the doc: 

> Note: This method sets the complete metadata of an object. You should generally use the PATCH method instead, which changes only metadata you define in the request, while leaving other metadata unaffected.

From: https://cloud.google.com/storage/docs/json_api/v1/objects/update

I'm open to support `update`, but it'd to be a new endpoint.

fsouza:
Closing for now, feel free to open a new PR with the implementation for update.

jonasfugedi:
Hello! I ran into a similar issue - running the  [java client](https://googleapis.dev/java/google-cloud-storage/latest/com/google/cloud/storage/Blob.html#update-com.google.cloud.storage.Storage.BlobTargetOption...-) a  POST request is sent (not the expected PUT as implied by the documentation). 

I tried the same trick of adding support for POST to the patch method as described in this PR, and that solves my use case. 

`r.Path("/b/{bucketName}/o/{objectName:.+}").Methods(http.MethodPatch, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.patchObject))`

I'm not a go programmer, but I'll have a look and see if I can add the update functionality...

Issue #214, opened 2020-04-16
Deleting a bucket fails


someone1:
First off, amazing project! Much needed.

When I try and delete a bucket to cleanup after my tests, it fails:

```go
	gcsTestBucketName := "test_bucket"
	transCfg := &http.Transport{
		// nolint:gosec // Purposely connecting to a local self-signed certificate endpoint
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true}, // ignore expired SSL certificates
	}
	httpClient := &http.Client{Transport: transCfg}
	client, err := storage.NewClient(ctx, option.WithEndpoint(os.Getenv("GCS_FAKE_SERVER")+"/storage/v1/"), option.WithHTTPClient(httpClient))
	if err != nil {
		t.Fatal(err)
	}

	if err = client.Bucket(gcsTestBucketName).Create(ctx, "test-project", nil); err != nil {
		t.Fatalf("could not create bucket; %v", err)
	}

	defer func() {
		if err = client.Bucket(gcsTestBucketName).Delete(ctx); err != nil {
			t.Errorf("could not delete bucket: %v", err)
		}
	}()
```

I end up getting a panic:

```
panic: runtime error: invalid memory address or nil pointer dereference [recovered]
        panic: runtime error: invalid memory address or nil pointer dereference
[signal SIGSEGV: segmentation violation code=0x1 addr=0x30 pc=0xc1854e]

goroutine 8 [running]:
testing.tRunner.func1.1(0xd66900, 0x1725600)
        /usr/lib/go/src/testing/testing.go:940 +0x2f5
testing.tRunner.func1(0xc0003ac900)
        /usr/lib/go/src/testing/testing.go:943 +0x3f9
panic(0xd66900, 0x1725600)
        /usr/lib/go/src/runtime/panic.go:969 +0x166
cloud.google.com/go/storage.(*BucketHandle).newDeleteCall(0xc000017cb8, 0xc000079d40, 0xeb5bc4, 0x29)
        /home/user/go/pkg/mod/cloud.google.com/go/storage@v1.6.0/bucket.go:107 +0x3e
cloud.google.com/go/storage.(*BucketHandle).Delete(0xc000017cb8, 0x10f9c40, 0xc000079d40, 0x0, 0x0)
        /home/user/go/pkg/mod/cloud.google.com/go/storage@v1.6.0/bucket.go:99 +0xed
// ...
```

fsouza:
[...]

PR #440, opened 2021-03-02
Add delete routes

kmgreen2:
I am not sure if others are hitting this, but I consistently got 405 whenever I tried issuing deletes using the official Golang GCS SDK.

Looking closer at the code, it seems like some of the routers are not in use?  In `buildMuxer()` (server.go:189), I see that `routers` is allocated, but never used.  Furthermore, I only see the routers defined on server.go:213 through server.go:225 when inspecting/debugging mux.go:136 (`mux:v1.8.0`).

What is the intention of `server.go:198`?  Should that be removed, or am I missing a side-effect.  If there is a side-effect associated with routers, I would suggest you directly define those routes.

In any case, deletes are returning a 405 because the server cannot find a route.  Since I am not sure of the intention of `server.go:198`, I have added new delete routes directly to `s.mux`, which fixes the issue.

### DIFF
--- a/examples/go/main.go
+++ b/examples/go/main.go
@@ -13,7 +13,9 @@ import (
 	"context"
 	"fmt"
 	"io/ioutil"
+    "io"
 	"log"
+    "bytes"
 
 	"cloud.google.com/go/storage"
 	"google.golang.org/api/iterator"
@@ -21,7 +23,7 @@ import (
 )
 
 func main() {
-	client, err := storage.NewClient(context.TODO(), option.WithEndpoint("http://localhost:8080/storage/v1/"))
+	client, err := storage.NewClient(context.TODO(), option.WithEndpoint("http://localhost:8080/storage/v1/"), option.WithoutAuthentication())
 	if err != nil {
 		log.Fatalf("failed to create client: %v", err)
 	}
@@ -40,6 +42,13 @@ func main() {
 		log.Fatal(err)
 	}
 	fmt.Printf("contents of %s/%s: %s\n", bucketName, fileKey, data)
+
+    payloadBytes := bytes.NewBuffer([]byte("foobar"))
+    err = uploadFile(client, bucketName, "uploadTestKey", payloadBytes)
+	if err != nil {
+		log.Fatal(err)
+	}
+	fmt.Printf("uploaded %s to %s/uploadTestKey\n", string(payloadBytes.Bytes()), bucketName)
 }
 
 func list(client *storage.Client, bucketName string) ([]string, error) {
@@ -65,4 +74,15 @@ func downloadFile(client *storage.Client, bucketName, fileKey string) ([]byte, e
 	}
 	defer reader.Close()
 	return ioutil.ReadAll(reader)
+}
+
+func uploadFile(client *storage.Client, bucketName, key string, reader io.Reader) error {
+    wc := client.Bucket(bucketName).Object(key).NewWriter(context.Background())
+	if _, err := io.Copy(wc, reader); err != nil {
+		return fmt.Errorf("io.Copy: %v", err)
+	}
+	if err := wc.Close(); err != nil {
+		return fmt.Errorf("Writer.Close: %v", err)
+	}
+	return nil
 }
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -215,7 +215,8 @@ func (s *Server) buildMuxer() {
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods("POST").HandlerFunc(s.insertObject)
 	s.mux.Path("/upload/resumable/{uploadId}").Methods("PUT", "POST").HandlerFunc(s.uploadFileContent)
 
-    s.mux.Host(bucketHost).Path("/b/{bucketName}/o/{objectName:.+}").Methods("DELETE").HandlerFunc(s.deleteObject)
+    s.mux.Path("/b/{bucketName}/o/{objectName:.+}").Methods("DELETE").HandlerFunc(s.deleteObject)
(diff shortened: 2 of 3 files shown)


fsouza:
Hi @kmgreen2, object deletion should work (bucket deletion is not implemented, see #214). The routers in the slice are defined as subrouters of the main router and that's where the side effect happens. Can probably add a comment or something like that, but that's unfortunately how gorilla/mux works.

Looking at your diff, it seems that your code is trying to delete objects without the `/storage/v1` prefix? Which SDK are you using? Can you share a reproducer?

fsouza:
Oh I can't read, you mentioned you were using the Go SDK. I can give it a shot later.

fsouza:
Can't dig into this right now, but opened #441 that should provide a reproducer.

If that PR passes, could you share some code to reproduce the behavior you're seeing?

kmgreen2:
@fsouza Thanks for the quick reply!

It is possible I am not starting the server properly (I just started using your library yesterday :) )

Here is how I am starting the GCS server (I had to hijack DNS resolution, so I can include this in tests without updating `/etc/hosts`):

```
func SetupGCSLocal() (*fakestorage.Server, error){
	return fakestorage.NewServerWithOptions(fakestorage.Options{
		InitialObjects: []fakestorage.Object{
			{
				BucketName: "localtest",
				Name:       "some/object/foo",
				Content:    []byte("bar"),
			},
		},
		Scheme: "http",
		Host: "localhost",
		Port: 8081,
		Writer: os.Stdout,
		PublicHost: "localtest.localhost",
	})
}
```

I then create a new client using the following.  It took some trial and error to get the SDK to work properly locally, which appears to be more of a Google SDK problem than anything to do with fake-gcs-server:

```
os.Setenv("STORAGE_EMULATOR_HOST", "localhost:8081")
storage.NewClient(ctx, 
    option.WithoutAuthentication(),
    option.WithHTTPClient(&http.Client{Transport: GetFakeTransport()}),
    option.WithEndpoint("http://localtest.localhost:8081"))
```

Thanks, again!

kmgreen2:
@fsouza I see that the tests do not rely on running the fake GCS daemon, could that be related?

kmgreen2:
I took a closer look at your examples and noticed that they only include list and download (and delete with your PR).  The code snippets above were definitely made much more complicated to support upload locally.  I tried to reproduce with upload and delete using `examples/`, but am unable to build the server locally or with Docker:

```
docker build -t fsouza/fake-gcs-server .
Sending build context to Docker daemon  275.5kB
Step 1/12 : FROM golang:1.16.0 AS builder
 ---> f15d23d96763
Step 2/12 : WORKDIR /code
 ---> Using cache
 ---> 31cf0090e68b
Step 3/12 : ADD go.mod go.sum ./
 ---> Using cache
 ---> 9e7876ffa5c1
Step 4/12 : RUN go mod download
 ---> Using cache
 ---> a65474170c25
Step 5/12 : ADD . ./
 ---> Using cache
 ---> 6d4d46dfb0f9
Step 6/12 : ENV CGO_ENABLED=0
 ---> Using cache
 ---> db56529fd654
Step 7/12 : RUN go build -o fake-gcs-server
 ---> Running in a1883d71be1b
package github.com/kmgreen2/fake-gcs-server
	main.go:18:2: use of internal package github.com/fsouza/fake-gcs-server/internal/config not allowed
The command '/bin/sh -c go build -o fake-gcs-server' returned a non-zero code: 1
```

In any case, my PR works for upload, download, delete and head, so I am unblocked by overloading go.mod with my branch.  Since this is just for tests, it should be fine for now.

kmgreen2:
Maybe part of my problem is that I am setting up my client to upload, download and delete, which requires more than your examples provide.  I added an upload step to your Go example and it does not work properly with the official Google Go SDK.

I added this helper:
```
func uploadFile(client *storage.Client, bucketName, key string, reader io.Reader) error {
    wc := client.Bucket(bucketName).Object(key).NewWriter(context.Background())
    if _, err := io.Copy(wc, reader); err != nil {
        return fmt.Errorf("io.Copy: %v", err)
    }
    if err := wc.Close(); err != nil {
        return fmt.Errorf("Writer.Close: %v", err)
    }
    return nil
}```

and called it after `downloadFile` in `example/go/main.go`:

```
   payloadBytes := bytes.NewBuffer([]byte("foobar"))
    err = uploadFile(client, bucketName, "uploadTestKey", payloadBytes)
    if err != nil {
        log.Fatal(err)
    }
    fmt.Printf("uploaded %s to %s/uploadTestKey\n", string(payloadBytes.Bytes()), bucketName)
```

I get this result:

```
$ go run main.go
buckets: [some_file.txt]
contents of sample-bucket/some_file.txt: Some amazing content to be loaded
2021/03/02 16:11:59 Writer.Close: Post "http://http/upload/storage/v1/b/sample-bucket/o?alt=json&name=uploadTestKey&prettyPrint=false&projection=full&uploadType=multipart": dial tcp: lookup http: no such host
exit status 1
```

Doing what I did above seems to work for upload, delete, download and head.  It seems that the official Google SDK doesn't like local instances of GCS, so requires a bit of extra configuration.

In any case, I am unblocked by using my PR and goofy client-side configuration, but would be happy to help if you think this is a general problem.

kmgreen2:
@fsouza FYI, I unblocked myself with this PR in my fork.  It took a bit of wrangling to get the Google SDK working for all endpoints locally.  If you think there is an actual problem here, I am happy to help, but understand if you don't.  I can just pin to my fork, since this represents a small number of tests in my project (https://github.com/kmgreen2/agglo/blob/main/pkg/storage/gcsobj_test.go) and I have them working with my fork.

PR #447, opened 2021-03-10
Increase time tolerance to avoid test flakes

drigz:
I observed this test failing 0.5% of the time on our CI server. After
this change, I ran the test 10,000 times without seeing any failures.

### DIFF
--- a/internal/backend/backend_test.go
+++ b/internal/backend/backend_test.go
@@ -211,9 +211,12 @@ func TestBucketCreateGetList(t *testing.T) {
 			if err == nil {
 				t.Fatalf("bucket %s, exists before being created", bucket.Name)
 			}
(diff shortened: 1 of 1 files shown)


drigz:
For context, here's an example test failure:

```
=== RUN   TestBucketCreateGetList/storage_backend_filesystem
    third_party/golang/fakegcsserver/internal/backend/backend_test.go:231: bucket {prod-bucket false 2021-03-09 19:53:37.999823884 -0800 PST} does not have the expected props after retrieving. Expected {prod-bucket false 0001-01-01 00:00:00 +0000 UTC} and time between 2021-03-09 19:53:38 -0800 PST and 2021-03-09 19:53:38.006195462 -0800 PST m=+0.111907208
--- FAIL: TestBucketCreateGetList (0.00s)
```

fsouza (approved):
Thanks for contributing!

Issue #451, opened 2021-03-15
Incorrect content type?


cgreening:
Thanks for the great library. I'm struggling to get it working with Rails - does anyone have this working?

I think one of the issues I'm seeing is caused by an incorrect content type stopping the ruby library from parsing the responses.

```
curl -i http://0.0.0.0:4443/storage/v1/b/test
HTTP/1.1 200 OK
Date: Mon, 15 Mar 2021 09:59:42 GMT
Content-Length: 111
Content-Type: text/plain; charset=utf-8

{"kind":"storage#bucket","id":"test","name":"test","versioning":{},"timeCreated":"2021-03-15T09:59:06.41438Z"}
```

cgreening:
I've hacked around with the code and added the correct content type and it fixes my issue with ruby not parsing the response. I'm not a go developer, but I'll try and pull together a pull request with a fix.

fsouza:
Oh interesting, I thought we had the proper content-types, but probably only on download. Should be an easy fix though. Let me know if you have any questions or would prefer me to fix it.

PR #452, opened 2021-03-15
Add application/json content type where required

cgreening:
This seems to be a bit horrible as I've had to add the header everywhere JSON is returned. I don't know if there's a better way to do this.

Fixes: https://github.com/fsouza/fake-gcs-server/issues/451

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -359,15 +359,16 @@ func (s *Server) listObjects(w http.ResponseWriter, r *http.Request) {
 	versions := r.URL.Query().Get("versions")
 
 	objs, prefixes, err := s.ListObjects(bucketName, prefix, delimiter, versions == "true")
-	w.Header().Set("Content-Type", "application/json")
 
 	encoder := json.NewEncoder(w)
 	if err != nil {
+		w.Header().Set("Content-Type", "application/json")
 		w.WriteHeader(http.StatusNotFound)
 		errResp := newErrorResponse(http.StatusNotFound, http.StatusText(http.StatusNotFound), nil)
 		encoder.Encode(errResp)
 		return
 	}
+	w.Header().Set("Content-Type", "application/json")
 	encoder.Encode(newListObjectsResponse(objs, prefixes))
 }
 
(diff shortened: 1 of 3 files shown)


fakestorage/object.go
```diff
@@ -421,6 +428,7 @@ func (s *Server) listObjectACL(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) setObjectACL(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
+	w.Header().Set("Content-Type", "application/json")
```
fsouza, line 431: can you move it to line 444?

fakestorage/object.go
```diff
@@ -408,6 +413,8 @@ func (s *Server) deleteObject(w http.ResponseWriter, r *http.Request) {
 }
 
 func (s *Server) listObjectACL(w http.ResponseWriter, r *http.Request) {
+	w.Header().Set("Content-Type", "application/json")
```
fsouza, line 416: can you move it to 425?

fakestorage/object.go
```diff
@@ -398,6 +401,8 @@ func (s *Server) getObject(w http.ResponseWriter, r *http.Request) {
 func (s *Server) deleteObject(w http.ResponseWriter, r *http.Request) {
 	vars := mux.Vars(r)
 	err := s.backend.DeleteObject(vars["bucketName"], vars["objectName"])
+	w.Header().Set("Content-Type", "application/json")
```
fsouza, line 404: move to 412

fsouza (commented):
Yeah, this is non-ideal, but if you move the header definitions to right before the response is encoded I can merge it to unblock you (I added inline comments to some examples, but there are other instances).

I can refactor it later to either use a helper function to write the response or wrap all handlers that can return JSON (pretty much all of them except for download? heh)

cgreening:
Let me know if that looks ok - we're not blocked as I can point our docker-compose at my fork for now.

fsouza:
> Let me know if that looks ok - we're not blocked as I can point our docker-compose at my fork for now.

No problem, I'll merge this and tag a release just to get it fixed. Hopefully I can find some time later this week to refactor the code. Thank you very much for contributing!

fsouza (approved):

PR #481, opened 2021-04-19
adds support for deleting buckets

joerodriguez:
This PR adds support for deleting buckets with the gcs client. I do not know the full set of preconditions for deleting a bucket, but GCS does require the bucket to be empty, which is respected here.

This fixes https://github.com/fsouza/fake-gcs-server/issues/214

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -107,6 +107,21 @@ func (s *storageFS) GetBucket(name string) (Bucket, error) {
 	return Bucket{Name: name, VersioningEnabled: false, TimeCreated: timespecToTime(createTimeFromFileInfo(dirInfo))}, err
 }
 
+// DeleteBucket removes the bucket from the backend.
+func (s *storageFS) DeleteBucket(name string) error {
+	objs, err := s.ListObjects(name, false)
+	if err != nil {
+		return BucketNotFound
+	}
+	if len(objs) > 0 {
+		return BucketNotEmpty
+	}
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	return os.RemoveAll(filepath.Join(s.rootDir, url.PathEscape(name)))
+}
+
 // CreateObject stores an object as a regular file in the disk.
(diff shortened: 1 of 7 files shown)


internal/backend/fs.go
```diff
@@ -107,6 +107,21 @@ func (s *storageFS) GetBucket(name string) (Bucket, error) {
...
+	if err != nil {
+		return BucketNotFound
+	}
+	if len(objs) > 0 {
+		return BucketNotEmpty
+	}
+
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+	return os.RemoveAll(filepath.Join(s.rootDir, url.PathEscape(name)))
```
joerodriguez, line 122: Any ideas for validating the name part of this path? I am using the same method for joining as ListObjects, so that may need validation as well. I could not find anything in the standard library that will securely join paths. There is a proposal, but with not much activity: https://pkg.go.dev/github.com/cyphar/filepath-securejoin

joerodriguez: @fsouza Actually, I am seeing just seeing the `validateBucketName` (https://github.com/fsouza/fake-gcs-server/blob/786e1fb8d78e66422ed0398290adac978ff66a52/fakestorage/bucket.go#L102) function now. It appears to be called before creating a bucket. It seems like we can also validate the bucket name before get, list objects, and delete. Does that sounds right? Even if we do that, I'm not sure that CodeQL will go green.

fsouza: @joerodriguez apologies for the unresponsiveness. We can either look into was to signal in the code that something is known to be safe or I can override it and merge. Once merged, I have an option to mark it as safe in the settings panel.

Alternatively, we could use the new fs interface, and that should be safe, but the problem is that it requires Go 1.16 and I want to support fake-gcs-server as a library too, where we support the last two versions of Go.

Either way, I haven't reviewed the PR yet. I'll review it later this week.

fsouza (approved):
Thank you for contributing!

Issue #346, opened 2020-10-19
Resumable upload fails with 400 using nodejs client. 


jeantil:
While working on #345 I was unable to make the default `file.save(...)` call work and was forced to add `{ resumable: false }` to make it pass. 

The error was: 
```
Error: Upload failed
    at Upload.<anonymous> (/home/jean/dev/startups/yupwego/src/fake-gcs-server/examples/node/node_modules/gcs-resumable-upload/build/src/index.js:191:30)
    at Upload.emit (events.js:327:22)
    at Upload.onResponse (/home/jean/dev/startups/yupwego/src/fake-gcs-server/examples/node/node_modules/gcs-resumable-upload/build/src/index.js:388:14)
    at Upload.makeRequestStream (/home/jean/dev/startups/yupwego/src/fake-gcs-server/examples/node/node_modules/gcs-resumable-upload/build/src/index.js:337:14)
    at processTicksAndRejections (internal/process/task_queues.js:97:5)
    at async Upload.startUploading (/home/jean/dev/startups/yupwego/src/fake-gcs-server/examples/node/node_modules/gcs-resumable-upload/build/src/index.js:219:13)
```

Running the script in a debugger yields an HTTP 400 error on the following request 
```
PUT https://[::]:4443/upload/resumable/c2bfc362c82655ce7c2c5074bd300d11
Content-Range: 'bytes 0-*/*', 
Authorization: 'Bearer ya29.c.KpMB4QfVy54ujAsnhGPSPjDDAGp…LPYC_aCWrzh57NuDr-91aMHax7j4d5nKTGdQFKC-T', User-Agent: 'google-api-nodejs-client/6.0.6', 
x-goog-api-client: 'gl-node/12.18.3 auth/6.0.6',
Accept: 'application/json'
```
result in 
```
400 Bad Request
'invalid Content-Range: bytes 0-*/*\n'
```

gnarea:
I ran into this issue, but instead of getting a `400`, the `file.save(...)` just hung. I fixed it by passing `resumable: false`.

gnarea:
@fsouza, #497 didn't fix this issue. It just made it different. Consider this:

```typescript
test('Bug', async () => {
  const bucketName = 'foo';
  const fileContents = Buffer.from('bar');

  const client = new Storage({
    apiEndpoint: 'http://127.0.0.1:8080',
    projectId: 'the id',
  });

  await client.createBucket(bucketName);

[...]

PR #497, opened 2021-05-18
Fix nodeJS file upload (issue #346)

dpoetzsch:
As https://github.com/fsouza/fake-gcs-server/issues/346 states, there is a bug when uploading files with the nodeJS client lib.

I debugged it and the problem is, that nodeJS sends `Content-Range: bytes 0-*/*` (see https://github.com/googleapis/gcs-resumable-upload/blob/master/src/index.ts#L456).

I could not really find a good documentation on the exact semantics but from observing the client lib I guess it means "start and end upload stream with all the data I provide you". I adjusted the code accordingly and added a test case. I also checked that the following node example works fine:

```javascript
async function saveFile() {
  // [START storage_list_buckets]
  // Imports the Google Cloud client library
  const { Storage } = require("@google-cloud/storage");

  // Creates a client
  const storage = new Storage({
    apiEndpoint: "http://localhost:8080",
    projectId: "test",
  });

  const bucket = await storage.createBucket("testbucket");
  const file = new File(bucket, "test.txt");
  await file.save("abc", { contentType: "text/plain" });
  // [END storage_list_buckets]
}

saveFile().catch((err) => {
  console.error(err);
  process.exit(1);
});
```

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -380,11 +380,7 @@ func (s *Server) uploadFileContent(r *http.Request) jsonResponse {
 			commit = parsed.KnownTotal && (parsed.End+1 >= parsed.Total)
 		} else {
 			// End of a streaming request
-			contentLength := 0
-			if obj.Content != nil {
-				contentLength = len(obj.Content)
-			}
-			responseHeader.Set("Range", fmt.Sprintf("bytes=0-%d", contentLength))
+			responseHeader.Set("Range", fmt.Sprintf("bytes=0-%d", len(obj.Content)))
 		}
 	}
 	if commit {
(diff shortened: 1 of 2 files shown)


fakestorage/upload.go
```diff
@@ -380,7 +380,11 @@ func (s *Server) uploadFileContent(r *http.Request) jsonResponse {
 			commit = parsed.KnownTotal && (parsed.End+1 >= parsed.Total)
 		} else {
 			// End of a streaming request
-			responseHeader.Set("Range", fmt.Sprintf("bytes=0-%d", len(obj.Content)))
+			contentLength := 0
+			if obj.Content != nil {
+				contentLength = len(obj.Content)
+			}
+			responseHeader.Set("Range", fmt.Sprintf("bytes=0-%d", contentLength))
```
fsouza, line 387: The new code is doing the same thing. `len` returns 0 for a nil slice: https://play.golang.org/p/CArFxLf4d4v

fsouza (commented):
Looks good, just one note about the response header.

fakestorage/upload.go
dpoetzsch: Thanks for the tip, I'm not yet that fluent in Go.

fsouza (approved):
Thanks!

Issue #392, opened 2021-01-11
Access control list in fake-gcs-server


azoom-l-v-long:
**First, thank you for writing this!**

**Currently, I want to use fake-gcs-server in a local development environment and t need some features for your company's projects.**

1. upload file: worked
2. copying, renaming, and moving objects: ……
3. get singed url: not working
4. using acl to restrict access level: not working

but for item 3, 4 does not work properly with my wishes
here, I use addAllUser
`const bucket = await storage.bucket('gcs-bucket-test')`
 `const acl = await bucket.file('fileName.png').acl`
`await acl.owners.addAllUsers()`
or
    `const options = {`
    `   entity: 'allUsers'`
    `   role: storage.acl.READER_ROLE`
    `}`
`const file = await bucket.file('fileName.png')`
    `await file.acl.add(options).then(function (data) {`
    `   const aclObject = data[0]`
    `   const apiResponse = data[1]`
    ` })`
**I received an error like below**
**Please let me know fake-gcs can assist me with this but this doesn't**
{ Error: Not found
    at new ApiError (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/@google-cloud/common/build/src/util.js:59:15)
    at Util.parseHttpRespBody (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/@google-cloud/common/build/src/util.js:194:38)
    at Util.handleResp (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/@google-cloud/common/build/src/util.js:135:117)
    at retryRequest (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/@google-cloud/common/build/src/util.js:434:22)
    at onResponse (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/retry-request/index.js:214:7)
    at res.text.then.text (/Users/azoomdev/Desktop/WORK/fake-gcs/node_modules/teeny-request/build/src/index.js:219:13)
    at process._tickCallback (internal/process/next_tick.js:68:7)
  code: 404,
  errors: null,
  response:
   PassThrough {
     _readableState:
      ReadableState {
        objectMode: false,
        highWaterMark: 16384,
        buffer: BufferList { head: null, tail: null, length: 0 },
        length: 0,
        pipes: null,
        pipesCount: 0,
        flowing: true,
        ended: true,
        endEmitted: true,
        reading: false,
        sync: false,
        needReadable: false,
[...]

PR #498, opened 2021-05-18
added POST endpoint for acls

dpoetzsch:
Besides the `PATCH` for object-level ACLs, there is also a `POST`: https://cloud.google.com/storage/docs/json_api/v1/objectAccessControls/insert

It is called for example by the nodeJS lib when calling `file.makePublic()`.

I added the endpoint and reused the `PATCH` code. I suspect the code is actually correct for `POST` but might be incomplete for `PATCH`. Anyhow, with this PR `file.makePublic()` works like a charm.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -192,6 +192,7 @@ func (s *Server) buildMuxer() {
 		r.Path("/b/{bucketName}/o").Methods("POST").HandlerFunc(jsonToHTTPHandler(s.insertObject))
 		r.Path("/b/{bucketName}/o/{objectName:.+}").Methods("PATCH").HandlerFunc(jsonToHTTPHandler(s.patchObject))
 		r.Path("/b/{bucketName}/o/{objectName:.+}/acl").Methods("GET").HandlerFunc(jsonToHTTPHandler(s.listObjectACL))
+		r.Path("/b/{bucketName}/o/{objectName:.+}/acl").Methods("POST").HandlerFunc(jsonToHTTPHandler(s.setObjectACL))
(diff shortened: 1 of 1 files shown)


fsouza:
@dpoetzsch thanks for contributing. fake-gcs-server doesn't support patch right now, only update: https://cloud.google.com/storage/docs/json_api/v1/objectAccessControls/update. It looks like the payload for update and insert are the same, so this should be compatible.

Could you write a test to confirm? If you're not familiar with the Go SDK, I can give it a try later.

dpoetzsch:
I'll give it a shot

dpoetzsch:
tl;dr: I think the Go Storage SDK does not use the POST endpoint which makes it difficult to confirm with a Go test I think.

If you search for `ObjectAccessControls.` in https://github.com/googleapis/google-cloud-go/blob/master/storage/acl.go you only find calls to `List`, `Update`, `Delete`.

I went on and also checked the request lib that is used underneath and actually found a definition for the POST endpoint: `DefaultObjectAccessControlsInsertCall` in https://github.com/googleapis/google-api-go-client/blob/master/storage/v1/storage-gen.go#L6051 However, this endpoint does not seem to be used at all.

The only option I see to write a test would be to call a nodeJS app from Go, because we know that `file.makePublic()` calls this endpoint. But this feels a bit over-complicated. What do you think?

dpoetzsch:
Also, I haven't checked, but this PR might also fix https://github.com/fsouza/fake-gcs-server/issues/392

fsouza:
@dpoetzsch That's fair enough. Unfortunately, this is not the first time we have to implement something because of how some SDKs behave. Thanks for contributing!

fsouza (approved):

PR #501, opened 2021-05-24
.net example

marcinbudny:
Added example showing how to work with the emulator from .net and C#


fsouza:
@marcinbudny thanks for contributing! Can you add the example to CI, to make sure we don't break it?

marcinbudny:
@fsouza Hi, I added the example to CI and dependabot. Testing this on a fork is a bit of a pain, but hopefully everything should work ok.

fsouza (approved):
Thanks!

Issue #313, opened 2020-08-26
Generate Crc32c and Md5Hash when seeding from disk


fsouza:
When loading objects from disk, we should also calculate the proper crc32 check and the md5 hash.

antonydenyer:
This is a broader topic than just calculating hashes.

Currently when using a preloaded data from disk data is fetched at execution time. ie when you call the `ListBuckets` endpoint the function will read from disk at the point in time. Subsequent calls to `ListBuckets` will just read from disk again. You could have the scenario where folders/buckets are created without going through the `createBucket` endpoint. I don't think this is a problem.

PR #504, opened 2021-05-27
Add CRC32 and MD5 hashes for preseeded files

IlyaM:
This patch adds CRC32 and MD5 hashes for files preloaded from the seed directory. I believe this closes the issue https://github.com/fsouza/fake-gcs-server/issues/313

I haven't noticed any preexisting tests for this functionality so I only tested the patch manually. Probably it makes sense to move CRC32 and MD5 functions out of upload.go into a separate file as they are now used in multiple places but I didn't do this to keep the patch small.


fsouza (approved):
I'll merge and move it to an internal package (the fakestorage package contains the public API used by libraries).

PR #510, opened 2021-06-11
Avoid integer overflow in large range requests

gaul:


fakestorage/object.go
```diff
@@ -544,7 +545,7 @@ func (s *Server) handleRange(obj Object, r *http.Request) (start, end int, conte
 				var err error
 				if end, err = strconv.Atoi(rangeParts[1]); err != nil {
 					end = len(obj.Content)
-				} else {
+				} else if end != math.MaxInt64 {
```
fsouza, line 548: For later, but we should probably use strconv.ParseInt instead of Atoi to ensure this is always 64 bits.

fsouza (approved):

PR #511, opened 2021-06-11
Always respond with HTTP 206 for range requests

gaul:
This matches the GCS behavior.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -516,8 +516,8 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	}
 
 	status := http.StatusOK
-	start, end, content := s.handleRange(obj, r)
-	if len(content) != len(obj.Content) {
+	ranged, start, end, content := s.handleRange(obj, r)
+	if ranged {
 		status = http.StatusPartialContent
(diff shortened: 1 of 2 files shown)


fakestorage/object.go
```diff
@@ -516,7 +516,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 
 	status := http.StatusOK
 	start, end, content := s.handleRange(obj, r)
-	if len(content) != len(obj.Content) {
+	if r.Header.Get("Range") != "" {
```
fsouza, line 519: Let's have `handleRange` return whether or not this is a "ranged" request, this way only that method is dealing with request headers for range.

gaul: Done.

fsouza (approved):

fsouza:
Change looks good. Can you fix the test?

gaul:
> Change looks good. Can you fix the test?

Done.

fsouza (approved):

gaul:
Combined with #510 now gcsfuse to read and write files.  fake-gcs-server integration will allow us to set up offline integration tests and CI.  Thanks!

PR #516, opened 2021-06-22
Fix some 32-bit incompatibilities

gaul:
This allows fake-gcs-server to work on Raspberry Pi.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -535,21 +535,21 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	}
 }
 
-func (s *Server) handleRange(obj Object, r *http.Request) (ranged bool, start, end int, content []byte) {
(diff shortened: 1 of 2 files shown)


fsouza (commented):
Nice catch. Do any of the existing tests fail when running in 32 bits? Thinking of adding it to CI to prevent regressions.

gaul:
Tests pass with this PR.  32-bit is a small niche these days but it usually doesn't take much effort to support.

fsouza:
> Tests pass with this PR.  32-bit is a small niche these days but it usually doesn't take much effort to support.

I know, I just want to make sure we have CI to prevent regressions. do the tests fail without the changes in this PR?

I can investigate some time this week.

fsouza:
ok confirmed that the code won't even compile on 386 (see #517). Will merge this, then rebase #517. Thanks for contributing!

fsouza (approved):

PR #528, opened 2021-07-06
Add support for startOffset and endOffset

### LINKED ISSUE #464: [Feature Request] add support for startOffset and endOffset
The GCP storage API allows specifying an inclusive startOffset and exclusive endOffset to fetch a range of objects. The fake storage server ignores both parameters, returning all objects instead.

https://cloud.google.com/storage/docs/json_api/v1/objects/list

lucaspin:
Closes #464, adding support for `startOffset` and `endOffset` when listing objects in a bucket

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -232,10 +232,26 @@ func (s *Server) createObject(obj Object) (Object, error) {
 	return fromBackendObjects([]backend.Object{newObj})[0], nil
 }
 
+type ListOptions struct {
+	Prefix      string
+	Delimiter   string
+	Versions    bool
+	StartOffset string
+	EndOffset   string
+}
+
 // ListObjects returns a sorted list of objects that match the given criteria,
 // or an error if the bucket doesn't exist.
-func (s *Server) ListObjects(bucketName, prefix, delimiter, startOffset, endOffset string, versions bool) ([]Object, []string, error) {
-	backendObjects, err := s.backend.ListObjects(bucketName, versions)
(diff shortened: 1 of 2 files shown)


fakestorage/object.go
```diff
@@ -234,7 +234,7 @@ func (s *Server) createObject(obj Object) (Object, error) {
 
 // ListObjects returns a sorted list of objects that match the given criteria,
 // or an error if the bucket doesn't exist.
-func (s *Server) ListObjects(bucketName, prefix, delimiter string, versions bool) ([]Object, []string, error) {
+func (s *Server) ListObjects(bucketName, prefix, delimiter, startOffset, endOffset string, versions bool) ([]Object, []string, error) {
```
fsouza, line 237: fake-gcs-server is also used as a library and this would be a breaking change. Can we introduce the feature without modifying the public API? Maybe create a method called `retrieveObjectsList` that takes all parameters you have currently and keep the signature of `ListObjects` the same (and `ListObjects` would just call `s.listObjects(bucketName, prefix, delimiter, "", "", versions)`). Or since the list of parameters is too long anyways, we could introduce a `ListObjectsWithOptions` that takes all the parameters as a struct. That would allow us to continue to evolve the API.

What do you think?

lucaspin: Hmmm, sorry, I didn't realize it was also used as a library. Anyway, I really like the idea of adding a new `ListObjectsWithOptions(bucketName, options)`, so I'll do that instead. Thanks for the suggestion!

lucaspin: Should be good now. Let me know what you think

fsouza (approved):
Thank you!

PR #530, opened 2021-07-12
Add file download to node example

### LINKED ISSUE #529: Can't download files with v5.8.2+ of the Node.js client
Version 5.8.1 of the Node.js client works fine, but versions 5.8.2 to 5.8.5 (latest) don't.

I suspect https://github.com/googleapis/nodejs-storage/pull/1419 is the culprit, but that must obviously be working against GCS servers.

Here's a simple repro script:

```typescript
test('Bug', async () => {
  const bucketName = 'foo';
  const fileContents = Buffer.from('bar');

  const client = new Storage({
    apiEndpoint: 'http://127.0.0.1:8080',
    projectId: 'the id',
  });

  await client.createBucket(bucketName);

  const file = client.bucket(bucketName).file('key');
  await file.save(fileContents, { resumable: false });

  const [fileRetrievalResponse] = await file.get();
  await expect(fileRetrievalResponse.download()).resolves.toEqual([fileContents]);
});
```

`fileRetrievalResponse.download()` fails with:

```
Error: The downloaded data did not match the data from the server. To be sure the content is the same, you should download the file again.
```

Server logs:

```
gcs_1    | time="2021-07-09T15:25:12Z" level=info msg="couldn't load any objects or buckets from \"/data\", starting empty"
gcs_1    | time="2021-07-09T15:25:12Z" level=info msg="server started at http://[::]:8080"
gcs_1    | time="2021-07-09T15:25:18Z" level=info msg="172.30.0.1 - - [09/Jul/2021:15:25:18 +0000] \"POST /storage/v1/b?project=the%20id HTTP/1.1\" 200 110"
gcs_1    | time="2021-07-09T15:25:18Z" level=info msg="172.30.0.1 - - [09/Jul/2021:15:25:18 +0000] \"POST /upload/storage/v1/b/foo/o?uploadType=multipart&name=key HTTP/1.1\" 200 357"
gcs_1    | time="2021-07-09T15:25:18Z" level=info msg="172.30.0.1 - - [09/Jul/2021:15:25:18 +0000] \"GET /storage/v1/b/foo/o/key? HTTP/1.1\" 200 400"
gcs_1    | time="2021-07-09T15:25:18Z" level=info msg="172.30.0.1 - - [09/Jul/2021:15:25:18 +0000] \"GET /storage/v1/b/foo/o/key?alt=media HTTP/1.1\" 200 3"
gcs_1    | time="2021-07-09T15:25:18Z" level=info msg="172.30.0.1 - - [09/Jul/2021:15:25:18 +0000] \"GET /storage/v1/b/foo/o/key? HTTP/1.1\" 200 400"
```

I've tested this against versions 1.21.2 and 1.28 of the `fake-gcs-server`.

aeaton-overleaf:
* Use `docker-compose.yaml` to start the `fake-gcs` service (copied from the `dotnet` example, with the addition of a `build` path and a `volumes` mount to load the example data).
* Add an example of downloading a file from a bucket.

The `download` example is currently failing when used with `@google-cloud/storage` > v5.8.1 unless `validation: false` is set, which isn't ideal - it seems that `fake-gcs` needs to be setting a `x-goog-hash` header on the response. 

I've added a line of code that sets `X-Goog-Hash` on the download response, but I'm not sure when either of `obj.Crc32c` or `obj.Md5Hash` are guaranteed to be set on the object, so someone with more knowledge of the server will need to finish this off.

Closes #529.

### DIFF
--- a/examples/node/index.js
+++ b/examples/node/index.js
@@ -17,11 +17,13 @@ async function run() {
   });
   // [END storage_list_buckets]
 
-  const response = await storage.bucket('sample-bucket')
+  const [content] = await storage.bucket('sample-bucket')
     .file('some_file.txt')
-    .download();
+    .download({
+      validation: false // FIXME
+    });
   console.log("Contents:")
-  console.log(response)
+  console.log(content.toString())
 }
 
 
--- a/examples/node/index.js
+++ b/examples/node/index.js
@@ -20,7 +20,7 @@ async function run() {
   const [content] = await storage.bucket('sample-bucket')
     .file('some_file.txt')
     .download({
-      validation: false // FIXME
+      // validation: false // FIXME: needed when x-goog-hash header isn't present on the response
     });
   console.log("Contents:")
   console.log(content.toString())
--- a/examples/node/index.js
+++ b/examples/node/index.js
@@ -1,4 +1,4 @@
-async function listBuckets() {
+async function run() {
   // [START storage_list_buckets]
   // Imports the Google Cloud client library
   const { Storage } = require("@google-cloud/storage");
@@ -16,9 +16,16 @@ async function listBuckets() {
     console.log(bucket.id);
   });
   // [END storage_list_buckets]
+
+  const response = await storage.bucket('sample-bucket')
+    .file('some_file.txt')
+    .download();
+  console.log("Contents:")
+  console.log(response)
 }
 
-listBuckets().catch((err) => {
+
+run().catch((err) => {
   console.error(err);
   process.exit(1);
 });
--- a/examples/node/index.js
+++ b/examples/node/index.js
@@ -19,9 +19,7 @@ async function run() {
 
   const [content] = await storage.bucket('sample-bucket')
     .file('some_file.txt')
-    .download({
-      // validation: false // FIXME: needed when x-goog-hash header isn't present on the response
-    });
+    .download();
   console.log("Contents:")
   console.log(content.toString())
 }
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -562,8 +562,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", strconv.Itoa(len(content)))
 	w.Header().Set(contentTypeHeader, obj.ContentType)
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
-	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s", obj.Crc32c))
-	// w.Header().Set("X-Goog-Hash", fmt.Sprintf("md5=%s", obj.Md5Hash))
+	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s,md5=%s", obj.Crc32c, obj.Md5Hash))
 	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
 	if obj.ContentEncoding != "" {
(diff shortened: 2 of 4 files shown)


fakestorage/object.go
```diff
@@ -562,6 +562,8 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", strconv.Itoa(len(content)))
 	w.Header().Set(contentTypeHeader, obj.ContentType)
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
+	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s", obj.Crc32c))
```
aeaton-overleaf, line 565: I'm not sure whether `obj.Crc32c` is generated by the server or comes from user input, so please check whether this value needs to be sanitized for use in the response header.

fakestorage/object.go
```diff
@@ -562,6 +562,8 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
...
+	// w.Header().Set("X-Goog-Hash", fmt.Sprintf("md5=%s", obj.Md5Hash))
```
fsouza, line 566: Do you have any links to the docs that say how google decides to which one to use? Or can it be either?

aeaton-overleaf: I haven't got anything definite, but I found this, which is mostly about sending the headers in upload requests but does say that "Cloud Storage stores MD5 hashes for all non-composite objects. CRC32Cs are available for all objects."

https://cloud.google.com/storage/docs/xml-api/reference-headers#xgooghash

fakestorage/object.go
fsouza: Sorry I missed this comment originally, but yeah those are generated on upload and when seeding the data, so we should be good!

fakestorage/object.go
aeaton-overleaf: That looks like what the Node client is expecting, yes:

https://github.com/googleapis/nodejs-storage/blob/390469a2283fd85ee3e1121d77505e77e16dbf10/src/file.ts#L1322-L1338

fsouza: Nice, can you update the code to do both?
```suggestion
	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s, md5=%s", obj.Crc32c, obj.Md5Hash))
```

aeaton-overleaf: I think the separator needs to be just a comma with no space, but yes apart from that. Does it need to handle cases where `obj.Crc32c` or `obj.Md5Hash` aren't set?

examples/node/index.js
```diff
@@ -16,9 +16,18 @@ async function listBuckets() {
     console.log(bucket.id);
   });
   // [END storage_list_buckets]
+
+  const [content] = await storage.bucket('sample-bucket')
+    .file('some_file.txt')
+    .download({
+      // validation: false // FIXME: needed when x-goog-hash header isn't present on the response
+    });
```
fsouza, line 24:
```suggestion
    .download();
```

fsouza (approved):
Thanks for contributing! I'll push a release once this is merged.

fakestorage/object.go
aeaton-overleaf: In the Node client it's possible to set `validation: 'crc32'` or `validation: 'md5'` as an option for the download, which might be a reason to include both in case one has been specified. I'm not sure how common that is, though.

fsouza:
I'll fix the dotnet issue on master.

PR #612, opened 2021-11-01
ci: build arm64 images

dllg:
Signed-off-by: Daniel Lundborg <daniel.lundborg@sandvik.com>

PR #639, opened 2021-12-28
ci: add multi arch build

### LINKED ISSUE #580: qemu segmentation fault on Apple M1 / arm64
The container crashes immediately after startup on an M1 Macbook. This is the log output:

fakegcs_1           | time="2021-09-23T18:41:16Z" level=info msg="server started at https://[::]:4443"
fakegcs_1           | qemu: uncaught target signal 11 (Segmentation fault) - core dumped

stoffeastrom:
Fixes #580  

Closes #612.

### DIFF
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -57,7 +57,6 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
-          file: ci/Dockerfile
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -30,7 +30,7 @@ jobs:
         id: meta
         uses: docker/metadata-action@v3
         with:
-          images: ${{ secrets.DOCKER_USERNAME }}/fake-gcs-server
+          images: fsouza/fake-gcs-server
           tags: |
             type=ref,event=branch
             type=ref,event=pr
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -57,6 +57,7 @@ jobs:
         uses: docker/build-push-action@v2
         with:
           context: .
+          file: ci/Dockerfile
           push: true
           tags: ${{ steps.meta.outputs.tags }}
           labels: ${{ steps.meta.outputs.labels }}
--- a/.github/workflows/docker-push.yml
+++ b/.github/workflows/docker-push.yml
@@ -3,35 +3,61 @@ on: create
 
 jobs:
   release:
+    if: github.event.ref_type == 'tag'
     runs-on: ubuntu-latest
     steps:
       - uses: actions/checkout@v2.4.0
-        if: github.event.ref_type == 'tag'
 
       - uses: actions/setup-go@v2.1.5
-        if: github.event.ref_type == 'tag'
+
         id: go
         with:
           go-version: 1.17
 
       - name: go-build
-        if: github.event.ref_type == 'tag'
+
         run: go build -o fake-gcs-server
         env:
           CGO_ENABLED: 0
 
       - name: sanity-check
-        if: github.event.ref_type == 'tag'
         uses: docker://alpine
         with:
           entrypoint: sh
           args: -c "./fake-gcs-server -h"
 
-      - name: docker-publish
-        if: github.event.ref_type == 'tag'
-        uses: fsouza/docker-publish-action@v1.0.1
+      - name: docker meta
+        id: meta
+        uses: docker/metadata-action@v3
         with:
-          dockerfile: ci/Dockerfile
-          docker-image: fsouza/fake-gcs-server
-          DOCKER-USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          docker-password: ${{ secrets.DOCKER_PASSWORD }}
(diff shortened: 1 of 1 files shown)


stoffeastrom:
Sample build https://github.com/stoffeastrom/fake-gcs-server/runs/4651907764?check_suite_focus=true#logs
Pushed it looks like https://hub.docker.com/layers/184587067/stoffeastrom/fake-gcs-server/latest/images/sha256-a6fb9cd8cf7c51ab6b75680b288054db65857e56dc4ff26b718a6506a2a36bfb?context=repo

image

.github/workflows/docker-push.yml
```diff
@@ -3,35 +3,44 @@ on: create
 
 jobs:
   release:
+    if: github.event.ref_type == 'tag'
```
fsouza, line 6: oh nice, thanks for fixing this 😁

.github/workflows/docker-push.yml
```diff
@@ -3,35 +3,44 @@ on: create
...
-          docker-image: fsouza/fake-gcs-server
-          DOCKER-USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          docker-password: ${{ secrets.DOCKER_PASSWORD }}
+          username: ${{ secrets.DOCKER_USERNAME }}
+          password: ${{ secrets.DOCKER_TOKEN }}
+      - name: build the image
+        run: |
+          docker buildx build \
+          --push \
+          --tag ${{ secrets.DOCKER_USERNAME }}/fake-gcs-server:latest \
```
fsouza, line 45: Nice, this looks good. I guess one thing missing would be the tagging scheme (for example, when we tag v1.2.3, we push `fsouza/fake-gcs-server:1`, `fsouza/fake-gcs-server:1.2` and `fsouza/fake-gcs-server:1.2.3`). 

Looks like docker provides another GH action that can accomplish this: https://github.com/docker/metadata-action#semver

I can take of it though, if you prefer. Up to you.

fsouza (commented):
Thank you very much for contributing and apologies for the slow turnaround time.

Just one note about tagging, bug looks good overall!

.github/workflows/docker-push.yml
stoffeastrom: Oh, yeah. totally forgot this. I'll have a look later tonight 👍

stoffeastrom:
with tagging https://github.com/stoffeastrom/fake-gcs-server/runs/4660625978?check_suite_focus=true
image

.github/workflows/docker-push.yml
```diff
@@ -3,35 +3,61 @@ on: create
...
+          password: ${{ secrets.DOCKER_TOKEN }}
+
+      - name: build and push
+        uses: docker/build-push-action@v2
         with:
-          dockerfile: ci/Dockerfile
-          docker-image: fsouza/fake-gcs-server
-          DOCKER-USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          docker-password: ${{ secrets.DOCKER_PASSWORD }}
+          context: .
```
stoffeastrom, line 59: Should we use `ci/Dockerfile` here instead?

stoffeastrom: we should remove `ci/Dockerfile` but maybe not in this pr

fsouza: Yeah I'll get rid of `ci/Dockerfile`. It was some premature optimization from before I used GH actions (I used to use `cloud.drone.io` where I could shave 10 seconds by running some stuff in parallel lols)

fsouza, line 59: Actually, can we keep using `ci/Dockerfile` for now? I'll revisit the build steps later.

```suggestion
          context: .
          file: ci/Dockerfile
```

fsouza (approved):
Thank you for contributing! 😁

.github/workflows/docker-push.yml
```diff
@@ -3,35 +3,62 @@ on: create
...
           args: -c "./fake-gcs-server -h"
 
-      - name: docker-publish
-        if: github.event.ref_type == 'tag'
-        uses: fsouza/docker-publish-action@v1.0.1
+      - name: docker meta
+        id: meta
+        uses: docker/metadata-action@v3
+        with:
+          images: ${{ secrets.DOCKER_USERNAME }}/fake-gcs-server
```
fsouza, line 33:
```suggestion
          images: fsouza/fake-gcs-server
```

DOCKER_USERNAME isn't actually fsouza 🙈  I use a bot account with limited privileges (I know it can push, but... 😬), just in case it leaks heh

stoffeastrom: :)

.github/workflows/docker-push.yml
```diff
@@ -3,35 +3,62 @@ on: create
...
+
+      - name: build and push
+        uses: docker/build-push-action@v2
         with:
-          dockerfile: ci/Dockerfile
-          docker-image: fsouza/fake-gcs-server
-          DOCKER-USERNAME: ${{ secrets.DOCKER_USERNAME }}
-          docker-password: ${{ secrets.DOCKER_PASSWORD }}
+          context: .
+          file: ci/Dockerfile
```
fsouza, line 60:
```suggestion
```

Sorry I'm wrong, we can't use `ci/Dockerfile` as it takes a prebuilt binary. If we do that we'll ship an amd64 binary in the arm64 image.

Sorry  for flip-flopping on you :(

stoffeastrom: np :D, this was what I meant but it got a bit confusing.. nvm, good catch before merging 👍

PR #659, opened 2022-01-25
Add "internal config update" endpoint to set externalUrl property after the fake server start

sergseven:
This change is dedicated to the problem with resumable uploads where it's necessary for fake-gcs-server to respond with "content" HTTP header with external URL which has to be accessible from storage client uses this fake server.

The suggested solution is to specify "external-url" at the endpoint command with the URL this fake server is accessible. 
It works for a regular case for local development, but it's getting more difficult when server is hosted on a container, (particularly using https://www.testcontainers.org/) so external URL is different than "public URL" fake-gcs-server" exposes.

Example:

```
@Testcontainers
class FakeGcsServerTest {
    // random port
    static int fakeGcsContainerPort = 4321;

    @Container
    static final GenericContainer<?> fakeGcs = new FixedHostPortGenericContainer<>("fsouza/fake-gcs-server")
        .withExposedPorts(4443)
        .withFixedExposedPort(4443, fakeGcsContainerIp)
        .withCreateContainerCmdModifier(cmd -> cmd.withEntrypoint(
            "/bin/fake-gcs-server",
            "-scheme", "http",
            // at this point there is no way to get a container IP
            // before the container started, so setting "localhost"
            "-external-url", "http://localhost:" + fakeGcsContainerPort
                                                                                                                 
        ));

    @Test
    void testWithFakeGcs() {
             var storageService = StorageOptions.newBuilder()
            .setHost("http://localhost:" + fakeGcs.getFirstMappedPort())
            .setProjectId("test-project")
            .setCredentials(NoCredentials.getInstance())
            .build()
            .getService();

            WriteChannel channel = storageService.writer(BlobInfo.newBuilder("sample-bucket2", "some_file2.txt").build());
            channel.write(ByteBuffer.wrap("line1\n".getBytes()));
            channel.write(ByteBuffer.wrap("line2\n".getBytes()));
            channel.close(); // <- ERROR on CI environment expected!
    }
```

The solution proposed is to modify fake-gcs-server configuration AFTER the container is eventually started and container IP is known:

```
String fakeGcsExternalUrl = "http://" + fakeGcs.getContainerIpAddress() + ":" + fakeGcs.getFirstMappedPort();
updateExternalUrlWithContainerUrl(fakeGcsExternalUrl);
```

See the full example at https://github.com/sergseven/fake-gcs-server/blob/upload-content-fix/examples/java/README.md#resumable-upload-operations-and-containerised-fake-gcs-server


fakestorage/server.go
```diff
@@ -220,6 +220,11 @@ func (s *Server) buildMuxer() {
 		r.Path("/b/{bucketName}/o/{objectName:.+}").Methods(http.MethodPut, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.updateObject))
 	}
 
+	// Internal / update server configuration
+	s.mux.Path("/internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
```
fsouza, line 224: nit: can we prefix it with an `_`. Not that relevant, just to make the intent even more explicit.
```suggestion
	s.mux.Path("/_internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
```

fakestorage/server.go
```diff
@@ -220,6 +220,11 @@ func (s *Server) buildMuxer() {
...
+	s.mux.Host(s.publicHost).Path("/internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
```
fsouza, line 225:
```suggestion
	s.mux.Host(s.publicHost).Path("/_internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
```

.tool-versions
```diff
@@ -0,0 +1 @@
+golang 1.17
```
fsouza, line 1: hmm I assume this is for asdf? Can you remove this file?

README.md
```diff
@@ -76,7 +76,7 @@ curl http://0.0.0.0:4443/storage/v1/b/sample-bucket/o
 
 ## Client library examples
 
-For examples using the Python, Node.js and Go clients, check out the
+For examples using the Python, Node.js, Java and Go clients, check out the
```
fsouza, line 79: Oh man this is outdated, can you include C# and Scala here?

fsouza: Or you can leave it alone and I'll generalize it so we don't have to remember to always add languages to the readme.

examples/go/main.go
```diff
@@ -78,3 +85,17 @@ func downloadFile(client *storage.Client, bucketName, fileKey string) ([]byte, e
 func deleteFile(client *storage.Client, bucketName, fileKey string) error {
 	return client.Bucket(bucketName).Object(fileKey).Delete(context.TODO())
 }
+
+func updateConfig() error {
+	changeExternalUrl := "http://localhost:8080/internal/config"
```
fsouza, line 90:
```suggestion
	changeExternalUrl := "http://localhost:8080/_internal/config"
```

examples/java/testcontainers/README.md
```diff
@@ -0,0 +1,62 @@
...
+        storageService = StorageOptions.newBuilder()
+            .setHost(fakeGcsExternalUrl)
+            .setProjectId("test-project")
+            .setCredentials(NoCredentials.getInstance())
+            .build()
+            .getService();
+    }
+
+    private static void updateExternalUrlWithContainerUrl(String fakeGcsExternalUrl) throws Exception {
+        String modifyExternalUrlRequestUri = fakeGcsExternalUrl + "/internal/config";
```
fsouza, line 27:
```suggestion
        String modifyExternalUrlRequestUri = fakeGcsExternalUrl + "/_internal/config";
```

examples/java/testcontainers/src/test/java/com/fsouza/fakegcsserver/java/examples/FakeGcsServerTest.java
```diff
@@ -0,0 +1,90 @@
...
+        storageService = StorageOptions.newBuilder()
+            .setHost(fakeGcsExternalUrl)
+            .setProjectId("test-project")
+            .setCredentials(NoCredentials.getInstance())
+            .build()
+            .getService();
+    }
+
+    private static void updateExternalUrlWithContainerUrl(String fakeGcsExternalUrl) throws Exception {
+        String modifyExternalUrlRequestUri = fakeGcsExternalUrl + "/internal/config";
```
fsouza, line 57:
```suggestion
        String modifyExternalUrlRequestUri = fakeGcsExternalUrl + "/_internal/config";
```

fsouza (commented):
Thanks for contributing! This looks good overall and is a good solution, I think we may be able to leverage it to solve other problems too.

I feel like the Java example is a bit overkill compared to others. But I understand you wanted to have an example of a solution for your problem. I guess that at some point I'll have to move those examples elsewhere, idk.

fakestorage/server.go
sergseven: looks good to me, renamed

fakestorage/server.go
sergseven: done

examples/go/main.go
sergseven: done

examples/java/testcontainers/README.md
sergseven: done

examples/java/testcontainers/src/test/java/com/fsouza/fakegcsserver/java/examples/FakeGcsServerTest.java
sergseven: done

README.md
sergseven: I would leave it as is.
Yep, I would recommend to have here a reference to the examples folder.

.tool-versions
sergseven: yes, asdf-related, removed

sergseven:
>I feel like the Java example is a bit overkill compared to others. But I understand you wanted to have an example of a solution for your problem. I guess that at some point I'll have to move those examples elsewhere, idk.

I agree, it would be good enough to only have a simple example without testcontainers, to give a good start for those who don't have to use testcontainers or containerise the fake-gcs-server.
Let me simplify the example to have just a no-container server started and simple Java client example.

examples/java/.mvn/wrapper/MavenWrapperDownloader.java
```diff
@@ -0,0 +1,117 @@
+/*
```
fsouza, line 1: qq: I haven't done Java in a while, are we supposed to check-in the .mvn folder or should we add it to `.gitignore`?

sergseven: Maven wrapper(.mvn, mvnw, mvnw.bat) is supposed to be checked in, so we don't have to add a requirement to have Maven distribution installed.

This is the best practice to stick to the specific maven version and have no requirement for Maven, only JDK is required then.

fsouza: That doesn't include the `.mvn` folder though, only `mvnw` and `mvnw.cmd`? Like, checking in a jar feels weird.

fsouza (approved):

examples/java/.mvn/wrapper/MavenWrapperDownloader.java
sergseven: `.mvn` actually contains wrapper/maven-wrapper.properties with maven version in there.
`maven-wrapper.jar` is optional to be checked in, so MavenWrapperDownloader.java is able to download this jar each time.

Nevertheless, I see you removed maven wrapper which I am completely fine with.
Thank you for review!

jonasfugedi:
Great work -- we really appreciate these improvements!

PR #666, opened 2022-01-31
Map errors to HTTP statuses

gaul:
The initial motivation is to map `ENAMETOOLONG` from 500 Internal Error
to 400 Bad Request.  This prevents clients like gcsfuse from
unnecessarily retrying requests.  This commit also remove some panic
call paths.

### DIFF
--- a/fakestorage/json_response.go
+++ b/fakestorage/json_response.go
@@ -2,7 +2,10 @@ package fakestorage
 
 import (
 	"encoding/json"
+	"errors"
 	"net/http"
+	"os"
+	"syscall"
 )
 
 type jsonResponse struct {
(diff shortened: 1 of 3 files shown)


gaul:
@fsouza Any better ideas on more precise error handling?

fakestorage/json_response.go
```diff
@@ -53,3 +55,12 @@ func (r *jsonResponse) getErrorMessage(status int) string {
 	}
 	return http.StatusText(status)
 }
+
+func errToJsonResponse(err error) jsonResponse {
+	// TODO: map more errno
```
fsouza, line 60: I think we can remove this TODO and just map errors on demand. What do you think?

fakestorage/json_response.go
```diff
@@ -53,3 +55,12 @@ func (r *jsonResponse) getErrorMessage(status int) string {
...
+	status := 0
+	if e, ok := err.(*os.PathError); ok && e.Err == syscall.ENAMETOOLONG {
```
fsouza, line 62: We probably want to use [errors.As](https://pkg.go.dev/errors#As), in case the error is wrapped.

fsouza (commented):
Thanks for contributing!

fakestorage/json_response.go
gaul: Done.

fakestorage/json_response.go
gaul: Done.

gaul:
CI was flaky when fetching a dependency.

fsouza (approved):

Issue #669, opened 2022-02-02
Efficiently handle large objects with filesystem backend


gaul:
Currently `multipartUpload` and `uploadFileContent` calls `loadContent` which calls `ioutil.ReadAll` which loads the entire request body into memory.  Instead it should stream the body to disk when using the filesystem backend.  This will reduce memory usage when testing large objects, e.g., 2 GB.

fsouza:
@gaul do you think we can close this now that #670 is merged and #671 is open to track the move away from JSON?

gaul:
These issues relate but have a somewhat different scope.  This one should include any memory optimization for large files, e.g., removing the call to `append(obj.Content)` in `uploadFileContent` which doubles memory usage.

PR #670, opened 2022-02-02
Stream JSON encoding and decoding to disk

gaul:
This reduces peak memory usage.  References #669.

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -134,11 +134,13 @@ func (s *storageFS) CreateObject(obj Object) (Object, error) {
 	if err != nil {
 		return Object{}, err
 	}
-	encoded, err := json.Marshal(obj)
+	file, err := os.OpenFile(filepath.Join(s.rootDir, url.PathEscape(obj.BucketName), url.PathEscape(obj.Name)), os.O_CREATE|os.O_WRONLY, 0o600)
 	if err != nil {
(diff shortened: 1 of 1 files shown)


internal/backend/fs.go
```diff
@@ -134,11 +134,13 @@ func (s *storageFS) CreateObject(obj Object) (Object, error) {
...
 	}
-	encoded, err := json.Marshal(obj)
+	file, err := os.OpenFile(filepath.Join(s.rootDir, url.PathEscape(obj.BucketName), url.PathEscape(obj.Name)), os.O_CREATE|os.O_WRONLY, 0o600)
 	if err != nil {
 		return Object{}, err
 	}
-	return obj, ioutil.WriteFile(filepath.Join(s.rootDir, url.PathEscape(obj.BucketName), url.PathEscape(obj.Name)), encoded, 0o600)
+	defer file.Close()
+	err = json.NewEncoder(file).Encode(obj)
+	return obj, err
```
fsouza, line 143: I assume that for encoding this doesn't matter? Agree it's a better pattern though.

fsouza (approved):

internal/backend/fs.go
gaul: Sorry I don't understand -- this should have the same encoding as `json.Marshal`?

PR #678, opened 2022-02-08
Store object metadata and data separately

### LINKED ISSUE #671: Store objects as binary instead of JSON
Storing objects in binary more efficient than encoding as JSON which has high memory usage for large files.  fake-gcs-server can store the metadata in extended attributes instead.  References #669.

gaul:
The former remains a JSON-encoded blob while the latter is stored in its
native format.  This reduces memory usage and is much faster by avoiding
JSON-encoding large objects.  This enables a future commit to avoid
reading the entire object, particularly for range requests.  Note that
this commit changes the on-disk format and is not compatible with
previous data sets.  References #669.  Fixes #671.

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -134,13 +134,23 @@ func (s *storageFS) CreateObject(obj Object) (Object, error) {
 	if err != nil {
 		return Object{}, err
 	}
-	file, err := os.OpenFile(filepath.Join(s.rootDir, url.PathEscape(obj.BucketName), url.PathEscape(obj.Name)), os.O_CREATE|os.O_WRONLY, 0o600)
+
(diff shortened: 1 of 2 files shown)


gaul:
@fsouza I believe this is the right thing to do but we should consider what kind of backwards-compatibility the project guarantees to the users.  This also might not be the optimal file format and storing metadata in extended attributes could be better.

fsouza:
Hey @gaul, thank you very much for contributing again! Do you think we'd run into any issues if we go for the solution based on extended attributes? Should we just go for it?

The thing I worry the most is Windows, but it's pure FUD: I have no idea what happens in the Windows world lol

gaul:
While Windows support xattr, the Go xattr library does not: pkg/xattr#47.  Personally I prefer to just go for it; see #683 for an alternate approach.

gaul:
Superseded by #683.

PR #683, opened 2022-02-12
Store object data natively and metadata in xattr

### LINKED ISSUE #671: Store objects as binary instead of JSON

gaul:
The latter remains a JSON-encoded blob but is now stored in file
extended attributes.  This reduces memory usage and is much faster by
avoiding JSON-encoding large objects.  This enables a future commit to
avoid reading the entire object, particularly for range requests.
Note that this commit changes the on-disk format and is not compatible
with previous data sets.  Extended attributes have some caveats
including lack of tmpfs and Windows support.  References pkg/xattr#47.
References #669.  Fixes #671


gaul:
This breaks Windows.  On one hand breaking existing (theoretical?) users is a bad thing.  On the other, the Linux xattr support only requires [141 lines](https://github.com/pkg/xattr/blob/master/xattr_linux.go) so a motivated user can add Windows support easily.

fsouza:
Perhaps we can introduce support for windows via json files for metadata and use build tags? What do you think?

gaul:
It would be nice if CI didn't immediately fail when one of the targets failed.

gaul:
Need some kind of `/var/tmp` workaround for Windows.  Linux needs to use this instead of `/tmp` to work around tmpfs xattr limitations.

fsouza:
> It would be nice if CI didn't immediately fail when one of the targets failed.

I think I can modify the config for that. I'll give it a shot later today.

internal/backend/xattr_linux.go
```diff
@@ -0,0 +1,27 @@
...
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
+package backend
+
+import (
+	"github.com/pkg/xattr"
+)
+
+const XATTR_KEY = "user.metadata"
```
fsouza, line 11: nit: let's call it `xattrKey`, this way it's not exported (I know it's an internal package, but still heh).
```suggestion
const xattrKey = "user.metadata"
```

internal/backend/xattr_linux.go
```diff
@@ -0,0 +1,27 @@
+// Copyright 2017 Francisco Souza. All rights reserved.
+// Use of this source code is governed by a BSD-style
+// license that can be found in the LICENSE file.
+
```
fsouza, line 4: Can you rename this file to `xatter_unix.go` and delete `xattr_darwin.go`? With the right build tags we should be able to do it:

```suggestion
//go:build !windows
// +build !windows

```

internal/backend/xattr_windows.go
```diff
@@ -0,0 +1,32 @@
...
+
+package backend
+
+import (
+	"io/ioutil"
+	"os"
+	"strings"
+)
+
+const XATTR_SUFFIX = ".metadata"
```
fsouza, line 16: Same as above regarding the naming convention
```suggestion
const xattrSuffix = ".metadata"
```

fsouza (commented):
Thank you very much for working this! Only some nits, but should be good to go otherwise.

internal/backend/xattr_linux.go
gaul: Done.

internal/backend/xattr_linux.go
gaul: Done.

internal/backend/xattr_windows.go
gaul: Done.

fsouza (approved):
Thank you very much!

Issue #684, opened 2022-02-13
uploaded file is crashed


ri8ika:
In mac machine, you can compare this issue by getting info of the file ("Get Info"). Under "More Info", there's no any value compared to original file. Additionally in "Preview", you can not see the preview (We can understand it is crashed). If the file is of text type, then the preview section will show the information in json format like metadata etc.

Now, let's analyze what happens on both mac and linux machines:

Mac machine:
When serving the file for the first time it loads the actual content. (Correct behavior)
After restarting the server and serving the file it loads the actual content. (Correct behavior)

Linux machine:
When serving the file for the first time it loads the actual content. (Correct behavior)
After restarting the server and serving the file it loads json content with the metadata etc information. (Wrong behavior)

This looks like the issue is similar to: https://github.com/fsouza/fake-gcs-server/issues/296

Even though I have mentioned Correct behavior, in fact is not correct behavior. Because file content is uploaded with some kind of json data (metadata, etc.) which is not correct.

But I have created this issue so that you guys can debug and solve the issue. Hope this issue helps!

fsouza:
Yeah this is related to us not persisting metadata properly I think. I believe the work @gaul is doing in #683 should fix this. My plan is to review that PR before the end of this week. It should be good to go though, as @gaul definitely know what he's doing. It's just a matter of me finding the time.

ri8ika:
Definitely. I have checked that as well and continuously watch the changes. If something is wrong (ie. result is not as expected) then I'll point out and let you know here.

By the way, I'm also debugging on this, if I could do something then I'd be glad to contribute.

ri8ika:
Ah, that's from different repo. I'll be checking that out later tonight and test it out.

ri8ika:
Hi @fsouza 

I tested #683 PR and it's working great! 

However, I have listed out some issues. Currently, I have tested in Mac machine only:

- Issue 1:
```
Step 1: Run project with docker, and upload file.
Step 2: user.metadata: operation not supported (xattr.Set)
```
- Issue 2:
```
When serving video (tested only with .mp4 format),
it's being downloaded instead of loading
```
- Issue 3:
```
Step 1: run project with go command
Step 2: uploaded file
Step 3: stop project
Step 4: run project with docker
uploaded file is there but it starts with empty,
that is to say uploaded file is not available while running with docker
```
- Issue 4:
```
When serving pdf file, and svg image:
It's not loading actual content, it's showing some encoded characters.
[...]

PR #685, opened 2022-02-16
Warn when rootDir lacks extended attribute support

gaul:
This hints to the user to reconfigure the server.  References #684.


fakestorage/server.go
```diff
@@ -102,6 +103,8 @@ type Options struct {
 
 	// Location used for buckets in the server.
 	BucketsLocation string
+
+	Logger *logrus.Logger
```
fsouza, line 107: hmm the thing to keep in mind is that fake-gcs-server is also used as a library, so we probably need a default logger that discards everything it receives or to wrap all calls to the logger with a nil check, which is probably not a great idea? 🤔

fsouza (approved):

PR #694, opened 2022-02-24
Write to byte buffer first, then to response

m4gus88:
I've ran into an issue where parts of the request could not be read, because the response writer was already close, as we were already writing stuff into it before.
The solution is to write into a byte buffer instead, and write that out in the end into the response.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -22,12 +22,12 @@ import (
 	"strings"
 	"sync"
 
-	"github.com/sirupsen/logrus"
 	"cloud.google.com/go/storage"
 	"github.com/fsouza/fake-gcs-server/internal/backend"
 	"github.com/fsouza/fake-gcs-server/internal/notification"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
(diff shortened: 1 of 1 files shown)


fsouza:
@m4gus88 not sure I follow. What's closing the ResponseWriter? Do you have a reproducer?

m4gus88:
I had an issue where after reading a few parts of the request, it couldn't read it anymore, because it got closed. Apparently, if you start writing into the response, it will get flushed eventually, and then the request reader will be closed as well.
The 2 fixes I've read were: either read the whole request first, and then start processing it, or write into a buffer first, and write that buffer out to the response in the end.

m4gus88:
@fsouza Reproducing it is difficult, because it's kinda random. For me it came up when I tried to download 5 documents in one batch. Sometimes it failed after reading 3 parts, sometimes 4 parts, sometimes it didn't fail at all. I think the flush/close happens in the background once you've written stuff in the response.

fsouza (approved):
One concern I had was about buffering all of that in memory, but I guess we can handle it.

Issue #332, opened 2020-09-19
Object metadata should be optional in resumable uploads


coryan:
When starting a resumable upload the body with object metadata is optional. And at least the C++ client library takes advantage of this.  The documentation is not explicit, but there are examples in the documentation that do not use a body:

https://cloud.google.com/storage/docs/performing-resumable-uploads#json-api

An equivalent request fails with `fake-gcs-server`:

```console
$ curl --verbose -X POST -H 'Content-Length: 0' -H 'Content-Type: application/json'  'http://localhost:4443/storage/v1/b/fake-bucket/o?uploadType=resumable&name=fake-object'
*   Trying ::1:4443...
* Connected to localhost (::1) port 4443 (#0)
> POST /storage/v1/b/fake-bucket/o?uploadType=resumable&name=fake-object HTTP/1.1
> Host: localhost:4443
> User-Agent: curl/7.72.0
> Accept: */*
> Content-Length: 0
> Content-Type: application/json
> 
* Mark bundle as not supporting multiuse
< HTTP/1.1 500 Internal Server Error
< Content-Type: text/plain; charset=utf-8
< X-Content-Type-Options: nosniff
< Date: Sat, 19 Sep 2020 13:21:25 GMT
< Content-Length: 4
< 
EOF
```

PR #699, opened 2022-03-01
Object metadata should be optional in resumable uploads #332

martinzink:
We are working on google-cloud integration for our project [Apache Nifi Minifi CPP](https://github.com/apache/nifi-minifi-cpp) and fake-gcs-server seems to be an awesome way to create integration tests.
However I encountered some issues, which made impossible to integrate fake-gcs-server with a C++ app

I already fixed the issues from google-cloud-cpp's side
https://github.com/googleapis/google-cloud-cpp/pull/8446
https://github.com/googleapis/google-cloud-cpp/pull/8477

The last issue that blocks using fake-gcs-server with a cpp application seems to be #332 
As described in the issue fake-gcs-server is stricter than google cloud storage when it comes to resumable upload requests.
With this PR the object metadata is optional when creating a resumable upload request.

I also added cpp examples so it can be verified that the resumable uploads are working now with fake-gcs-server.


examples/cpp/list_bucket.cpp
```diff
@@ -0,0 +1,15 @@
+#include "google/cloud/storage/client.h"
```
fsouza, line 1: Can you add this to CI? 

For that we need:

1. A script in `./ci`
2. A docker image where that script can run (probably something with cmake preinstalled?)
3. Then add it to the matrix here: https://github.com/fsouza/fake-gcs-server/blob/cf3fcb083e19553636419818e29f84825bd1e13c/.github/workflows/main.yml#L120-L139

fakestorage/upload.go
```diff
@@ -327,10 +327,7 @@ func (s *Server) multipartUpload(bucketName string, r *http.Request) jsonRespons
 func (s *Server) resumableUpload(bucketName string, r *http.Request) jsonResponse {
 	predefinedACL := r.URL.Query().Get("predefinedAcl")
 	contentEncoding := r.URL.Query().Get("contentEncoding")
-	metadata, err := loadMetadata(r.Body)
-	if err != nil {
```
gaul, line 331: Can you add a test that demonstrates this behavior?

Also is it possible to retain this check if we narrow it to ignore specific kinds of errors, e.g., missing metadata?

examples/cpp/list_bucket.cpp
fsouza: Oh 3 minutes isn't too bad given that CI runs things in parallel and running the tests on Windows takes 6 minutes anyway. We could definitely optimize it later though (would probably require maintaining a custom image, or leverage some caching mechanism in GitHub Actions).

fsouza (approved):
Thanks for contributing and sorry for the delayed review!

PR #710, opened 2022-03-09
Add possibility to provide your own certificates to the server

### LINKED ISSUE #709: Use own certificates when using SSL
when connecting through https the default certificate is only valid for example.com and 127.0.0.1. when using fake-gcs-server in a docker compose and naming it differently the certificates are always invalid. 
when using this from java it is really hard to use in a insecure matter.

I will create a PR to add you own certificates

krisgeus:
Closes #709.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -9,6 +9,7 @@ import (
 	"bytes"
 	"compress/gzip"
 	"context"
+	"crypto/tls"
 	"fmt"
 	"io"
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Nice, thanks for contributing! Can you fix the build failure? Then I can merge it.

krisgeus:
Fixed the linting error.

fsouza (approved):
Thanks for contributing!

Issue #711, opened 2022-03-10
Usage with testContainers-go


majelbstoat:
I'm trying to programmatically start fake-gcs-server using testcontainers-go. Example:

```go
       // Start a container listening on port 3000 internally.
        storagePort := 3000
	container, err := testcontainers.GenericContainer(ctx, testcontainers.GenericContainerRequest{
		ContainerRequest: testcontainers.ContainerRequest{
			Image:          "fsouza/fake-gcs-server:latest",
			Entrypoint:     []string{"/bin/fake-gcs-server", "-data", "/data", "-scheme", "http"},
			BindMounts:     map[string]string{"/data": e.hostDir},
			ExposedPorts:   []string{TCPPort(storagePort)},
			WaitingFor:     wait.ForLog("server started at"),
			Networks:       []string{Network},
			NetworkAliases: map[string][]string{Network: {StorageHostname}},
		},
		Started: true,
	})
	if err != nil {
		return err
	}

        // map this internal port to a randomly chosen external port.
	port, err := container.MappedPort(ctx, storagePort)
	if err != nil {
		return err
	}

        // set the emulator host and create a client.
        os.Setenv("STORAGE_EMULATOR_HOST", fmt.Sprintf("localhost:%s", port.Port())
        client, err  := storage.NewClient()
}
```

As with other folks, I can `curl --insecure http://localhost:XXXX/storage/v1/b/sample-bucket/o` and see my files in there, where `XXXX` is the mapped external port, but I can't retrieve them using the storage client. 

From other issues, I believe this is because I'm not setting a public host. My problem is, I don't know what the full public host will be until after the container starts, because that's when I get the mapped port.

As others have note, `-public-host 0.0.0.0` and `-public-host localhost` don't work, so I have this circular dependency which I can't figure out how to break.

fsouza:
@sergseven recently added an internal endpoint for updating some configuration parameters, aimed specifically at setting the external URL after starting the container with test containers, you can see a Java example in his PR: https://github.com/fsouza/fake-gcs-server/pull/659#issue-1114081250

Let me know if that's enough information is enough for you.

We should probably document that endpoint in the README at some point.

majelbstoat:
So, that looks like it might be helpful, thanks! But, at the moment, it looks like it can only be used to change the ExternalUrl.

https://github.com/fsouza/fake-gcs-server/blob/main/fakestorage/config.go

I can submit a patch for changing the public host, if that's acceptable?

[...]

PR #712, opened 2022-03-10
enable updating of publicHost through server config

majelbstoat:
Potentially fixes https://github.com/fsouza/fake-gcs-server/issues/711


fsouza:
> Potentially fixes #711

Would you be able to build a local image and test it locally with testContainers-go? (sorry not totally familiar with that library)

fsouza:
Yeah looking at the code I think this won't be enough because the publicHost is used once during the server initialization: https://github.com/fsouza/fake-gcs-server/blob/71fdd6b2ac7a2a3dc491a983129bb2d9312b5b72/fakestorage/server.go#L215-L269

I think we'll need to use [gorilla's MatcherFunc](https://pkg.go.dev/github.com/gorilla/mux#Route.MatcherFunc) instead to capture the server instance in a closure and check s.publicHost there.

majelbstoat:
Yeah, I was just doing that and unfortunately, this change alone isn't going be sufficient, because the server and muxer won't pick it up. I'll look into how to do that.

fakestorage/server.go
```diff
@@ -268,6 +268,18 @@ func (s *Server) buildMuxer() {
...
+// publicHostMatcher matches incoming requests against the currently specified server publicHost.
+func (s *Server) publicHostMatcher(r *http.Request, rm *mux.RouteMatch) bool {
+	return r.Host == s.publicHost
+}
+
+// bucketHostMatcher matches incoming requests against the bucket subdomain of the
+// currently specified server publicHost.
+func (s *Server) bucketHostMatcher(r *http.Request, rm *mux.RouteMatch) bool {
+	bucketHost := fmt.Sprintf("{bucketName}.%s", s.publicHost)
+	return r.Host == bucketHost
```
fsouza, line 280: Should this be `return r.Host == bucketHost || r.host == s.publicHost`? Because we want `storage.googleapis.com` and `<bucket>.storage.googleapis.com` to work?

fsouza: (and then whatever the user sets it to)

majelbstoat: i actually just removed that method. this was a direct replacement for the just the bucket subdomain routes. each one of those routes already had an equivalent non-subdomain route.

however, one interesting this is that tests failed when i converted the subdomain routes. the tests that test those routes actually suffer from the same problem as this fixes for use with containers: the test servers are defined like `NewServerWithOptions(Options{PublicHost: "127.0.0.1"})`, missing the port, because the port is only known after the httptest server has started.

fakestorage/server.go
```diff
@@ -252,10 +252,10 @@ func (s *Server) buildMuxer() {
 	s.mux.Path("/upload/resumable/{uploadId}").Methods(http.MethodPut, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.uploadFileContent))
 
 	// Batch endpoint
-	s.mux.Host(s.publicHost).Path("/batch/storage/v1").Methods(http.MethodPost).HandlerFunc(s.handleBatchCall)
+	s.mux.MatcherFunc(s.publicHostMatcher).Path("/batch/storage/v1").Methods(http.MethodPost).HandlerFunc(s.handleBatchCall)
 	s.mux.Path("/batch/storage/v1").Methods(http.MethodPost).HandlerFunc(s.handleBatchCall)
 
-	s.mux.Host(s.publicHost).Path("/{bucketName}/{objectName:.+}").Methods(http.MethodGet, http.MethodHead).HandlerFunc(s.downloadObject)
+	s.mux.MatcherFunc(s.publicHostMatcher).Path("/{bucketName}/{objectName:.+}").Methods(http.MethodGet, http.MethodHead).HandlerFunc(s.downloadObject)
```
fsouza, line 258: Could we extend one of the tests for these routes to test this scenario where the publicHost changes after the server is instantiated? (no need to invoke the internal config endpoint, we can simply mutate it directly in the test)

fsouza (commented):
Thank you very much for contributing, just one question about a test to check the behavior.

fakestorage/server.go
majelbstoat: I added a new test, TestDownloadAfterPublicHostChange.

fsouza (approved):

miklosboros:
I think this PR introduced a change in route matching behaviour which broke a couple of implementations (amongst them ours). The original host matcher from gorilla/mux: [original host matcher](https://github.com/gorilla/mux/blob/master/route.go#L293) enabled adding only the host (without port) as the public-host (and still matching on it) when starting the fake server, example: 
`/bin/fake-gcs-server -data /data -scheme http -public-host localhost`
while the new publicHostMatcher only allows it if the port number is also passed in with -public-host.

This seems to be because the new matcher is looking for an exact match, while the original one used a regexp matcher (from the documentation: "{name} matches anything until the next dot.").

It can be that this new behaviour is the only originally intended one too, however I wanted to point this out.

fsouza:
Oh interesting, I didn't realize gorilla/mux treated the string as a regexp. Thank you very much for pointing that out!

I guess we can fix that to keep the previous behavior. I'll send a change and make sure we have a test to prevent regressions in the future.

majelbstoat:
Sorry about the trouble, folks 😬

PR #728, opened 2022-03-25
server: make publicHostMatcher backwards compatible with gorilla/mux

### LINKED ISSUE #713: Failed to get object with error `storage: object doesn't exist` @ `1.37.0`
Hi, thank you for the work. However, when upgrade to `v1.37.0`, `object` was failed to be retrieved by gcs client. The tricky part is the object does exist and accessible by `curl`.

```
curl -I --insecure https://0.0.0.0:4443/storage/v1/b/foo/o/bar.txt | jq .

{
  "kind": "storage#object",
  ...
}
```

```go
func setup() (gcsClient *storage.Client) {
	transCfg := &http.Transport{
		TLSClientConfig: &tls.Config{InsecureSkipVerify: true},
	}
	httpClient := &http.Client{Transport: transCfg}
	gcsClient, _ = storage.NewClient(
		context.TODO(),
		option.WithEndpoint("https://localhost:4443/storage/v1/"),
		option.WithHTTPClient(httpClient),
	)
	return
}

func main() {
	gcsClient := setup()
	obj := gcsClient.Bucket("foo").Object("bar.txt").ReadCompressed(true)

	_, err := obj.NewReader(context.Background())
	if err != nil {
		log.Fatal(err.Error()) // storage: object doesn't exist
	}
}
```

### Version

- cloud.google.com/go/storage:  v1.16.0
- fake-gcs-server: 1.37.0

fsouza:
Fixes #713.


fsouza:
@miklosboros @chunming-c wanna give this a shot? I verified that `TestDownloadPartialPublicHostMatch` passes if we revert the changes in server.go (the other tests fail because they're testing the new behavior), but always good to double check before merging the PR.

fakestorage/server.go
```diff
@@ -270,7 +271,11 @@ func (s *Server) buildMuxer() {
 
 // publicHostMatcher matches incoming requests against the currently specified server publicHost.
 func (s *Server) publicHostMatcher(r *http.Request, rm *mux.RouteMatch) bool {
-	return r.Host == s.publicHost
+	if strings.Contains(s.publicHost, ":") {
+		return r.Host == s.publicHost
+	}
+	matched, _ := regexp.MatchString("^"+regexp.QuoteMeta(s.publicHost), r.Host)
```
fsouza, line 277: doesn't look like the right thing here. Will fix to make sure we don't match hosts partially either (will be annoying to test though).

Issue #299, opened 2020-08-05
Support for preconditions


charleskorn:
Does this server support preconditions? (https://cloud.google.com/storage/docs/generations-preconditions#_Preconditions)

I've tried adding the following to my upload code:

```go
w := c.bucket.
	Object("myobject.json").
	If(cloudstorage.Conditions{DoesNotExist: true}).
	NewWriter(ctx)
```

But the server always overwrites the object, even if it exists, instead of returning a HTTP 412 response. I can see that the upload URL contains `ifGenerationMatch=0` in the server's logs.

someone1:
Yikes, the `v.1.20.0` release was a breaking one - all my tests started to fail. Glad to see some progress being made on this but maybe we can keep the original behavior of allowing the request through vs a 501 response?

someone1:
A quick follow-up: was the implementation in #300 the correct placement for such a feature? The generation feature seems backend dependent (e.g. the memory backend supports file generations whereas the file backend does not). Would it have made more sense to expand the `Storage` interface with a `CreateObjectIfGenerationMatch` or something to be used instead of `CreateObject` as needed, or expand `CreateObject` to take in a precondition struct to process? This way, each backend can properly implement the features it supports (with locking) vs the current approach.

Thoughts?

fsouza:
Ugh, sorry about that :(( We should expand our shared tests to cover more behaviors. I'll make sure to keep an eye for that in the future (most of my use cases are using the memory backend :/).

I feel like extending the backend's CreateObject method is the best approach here, the same way we extended CreateBucket to pass versioningEnabled.

bmansoob:
👋 @fsouza 

I see generation support has been added in https://github.com/fsouza/fake-gcs-server/pull/735 at the same layer as in #300.

I added support for this in the memory backend [here](https://github.com/fsouza/fake-gcs-server/compare/main...bmansoob:generation-match-support?expand=1). I wanted to confirm with you first, before opening a pull request, that do you still plan on tackling this in a backend specific way.

PR #735, opened 2022-04-04
upload: add support for precondition headers

tymeshifter:
Added support for:

- ifGenerationMatch
- ifGenerationNotMatch

Without this support I am experiencing issues described in https://github.com/googleapis/google-cloud-go/issues/1624#issuecomment-1081701913

Relates to https://github.com/fsouza/fake-gcs-server/issues/299

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -150,17 +150,54 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
 func (s *Server) checkUploadPreconditions(r *http.Request, bucketName string, objectName string) *jsonResponse {
 	ifGenerationMatch := r.URL.Query().Get("ifGenerationMatch")
 
-	if ifGenerationMatch == "0" {
-		if _, err := s.backend.GetObject(bucketName, objectName); err == nil {
+	if ifGenerationMatch != "" {
+		gen, err := strconv.ParseInt(ifGenerationMatch, 10, 64)
+		if err != nil {
+			return &jsonResponse{
+				status:       http.StatusBadRequest,
+				errorMessage: err.Error(),
+			}
+		}
+		_, err = s.backend.GetObjectWithGeneration(bucketName, objectName, gen)
+		if gen == 0 {
+			if err == nil {
+				return &jsonResponse{
+					status:       http.StatusPreconditionFailed,
+					errorMessage: "Precondition failed",
+				}
+			}
+		} else if err != nil {
 			return &jsonResponse{
 				status:       http.StatusPreconditionFailed,
 				errorMessage: "Precondition failed",
 			}
 		}
-	} else if ifGenerationMatch != "" || r.URL.Query().Get("ifGenerationNotMatch") != "" {
-		return &jsonResponse{
-			status:       http.StatusNotImplemented,
-			errorMessage: "Precondition support not implemented",
+		return nil
+	}
+
+	ifGenerationNotMatch := r.URL.Query().Get("ifGenerationNotMatch")
+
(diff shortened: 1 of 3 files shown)


drGeoBN:
LGTM

fakestorage/upload.go
```diff
@@ -150,17 +150,54 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
...
 			return &jsonResponse{
 				status:       http.StatusPreconditionFailed,
 				errorMessage: "Precondition failed",
 			}
 		}
-	} else if ifGenerationMatch != "" || r.URL.Query().Get("ifGenerationNotMatch") != "" {
-		return &jsonResponse{
-			status:       http.StatusNotImplemented,
-			errorMessage: "Precondition support not implemented",
+		return nil
```
fsouza, line 175: random/non-blockign q: does this match the behavior of GCS when both parameters are provided. I know that providing both doesn't make sense, just wanted to confirm GCS prioritizes `ifGenerationMatch` in the presence of both (or maybe it fails with 400 Bad Request? idk)

fsouza (approved):
Thanks for contributing!

fakestorage/upload.go
```diff
@@ -150,17 +150,54 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
...
+		gen, err := strconv.ParseInt(ifGenerationMatch, 10, 64)
+		if err != nil {
+			return &jsonResponse{
+				status:       http.StatusBadRequest,
+				errorMessage: err.Error(),
+			}
+		}
+		_, err = s.backend.GetObjectWithGeneration(bucketName, objectName, gen)
+		if gen == 0 {
+			if err == nil {
```
orsinium, line 163: Shouldn't it be `if err != nil` as it used to be before? With this change, [my code](https://github.com/orsinium-labs/dimutex/blob/master/dimutex/_gcs.py#L117-L120) started to get `OK` status where a `Precondition Failed` status is expected and I'm trying to find out why. This line is my primary suspect.

fsouza: Hmm, I don't think so. ifGenerationMatch=0 means that the object must not exist. `err == nil` means the object exists, so that's why this path fails with `precondition failed`.

Do you have a reproducer or a request dump?

orsinium: I've also checked. You're right, changing this line doesn't help.

I linked the code of the library I use. If I don't find the cause myself, I'll open an issue with steps to reproduce. Thank you for a quick response.

fsouza: Oh, I guess the difference from before is that we're invoking `GetObjectWithGeneration` instead of `GetObject`, and that fails with the filesystem backend, because it doesn't support generations currently. We should really implement generations on that backend.

Is using the memory backend an option for you? Or are you dealing with large objects? I guess we could special-case ifGenerationMatch=0 to fix the breaking change while the filesystem backend doesn't support it. Any thoughts?

orsinium: Confirmed, using `-backend memory` fixes the bug. For me, in-memory backend should work but it's good to have it fixed. I'll create an issue for it, I already made a small Python script to reproduce the bug.

orsinium: Ah, I see #744. Thank you. I'll just comment there.

fsouza: Yeah implementing generation support in the filesystem backend is going to take some work, but I'm working on a quickfix for when ifGeneration=0 to keep it backwards compatible. I should have a PR very soon.

Issue #744, opened 2022-04-08
Support for generations in the filesystem backend


fsouza:
Currently, only the memory backend supports generations, we should implement generations in the filesystem backend too.

One thing to notice is that while we don't have any promises in terms backwards compatibility for the contents of `-filesystem-root`, maybe we shouldn't break it too badly? Meaning, we'd probably want to introduce a new backend, or some migration script (I doubt users would notice the migration script until too late? idk).

PR #745, opened 2022-04-08
upload: restore behavior when ifGenerationMatch=0 for the fs backend

fsouza:
This change is a follow-up to #735: previously, ifGenerationMatch=0 was
special-cased and the server used GetObject, which works on filesystem
and memory backends. That pull request changed the code to use
GetObjectWithGeneration, which is not supported by the filesystem
backend.

We couldn't catch this in tests because the test suite only runs with
the memory backend, so as part of this change I'm also including the
ability to use the filesystem storage in the test suite. That can get
slow though.

This is a hack and should be revisited once we implement #744.

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -160,15 +160,15 @@ func (s *Server) checkUploadPreconditions(r *http.Request, bucketName string, ob
 				errorMessage: err.Error(),
 			}
 		}
-		_, err = s.backend.GetObjectWithGeneration(bucketName, objectName, gen)
 		if gen == 0 {
+			_, err := s.backend.GetObject(bucketName, objectName)
 			if err == nil {
 				return &jsonResponse{
 					status:       http.StatusPreconditionFailed,
 					errorMessage: "Precondition failed",
 				}
 			}
-		} else if err != nil {
+		} else if _, err := s.backend.GetObjectWithGeneration(bucketName, objectName, gen); err != nil {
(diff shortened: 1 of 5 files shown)


fsouza:
@orsinium can you confirm that this works for you? If so, I can go ahead and merge this PR, then tag a new release.

fakestorage/upload.go
```diff
@@ -160,15 +160,15 @@ func (s *Server) checkUploadPreconditions(r *http.Request, bucketName string, ob
...
 		if gen == 0 {
+			_, err := s.backend.GetObject(bucketName, objectName)
 			if err == nil {
 				return &jsonResponse{
 					status:       http.StatusPreconditionFailed,
 					errorMessage: "Precondition failed",
 				}
 			}
-		} else if err != nil {
+		} else if _, err := s.backend.GetObjectWithGeneration(bucketName, objectName, gen); err != nil {
```
fsouza, line 171: The diff looks much larger because I had to build the capability to use the filesystem backend in tests, but this is the meat of it.

orsinium:
I'm late to the party :) I see it's already merged. Do you still need my input? You can also try to run the script I attached to the issue, it reproduces the bug.

fsouza:
> I'm late to the party :) I see it's already merged. Do you still need my input? You can also try to run the script I attached to the issue, it reproduces the bug.

Yes, looks like we're good!

PR #778, opened 2022-05-09
Content-Range is incorrectly specified.

yuki2006:
For example, a file of 22709217 bytes content is

should be `0-22709216/22709217` instead of `0-22709217/22709217`.

Chrome will not handle this correctly and videos and other files will not play.

image

RFC:
https://httpwg.org/specs/rfc7233.html#header.content-range

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -648,10 +648,10 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	}
 
 	status := http.StatusOK
-	ranged, start, end, content := s.handleRange(obj, r)
+	ranged, start, lastByte, content := s.handleRange(obj, r)
 	if ranged {
 		status = http.StatusPartialContent
-		w.Header().Set("Content-Range", fmt.Sprintf("bytes %d-%d/%d", start, end, len(obj.Content)))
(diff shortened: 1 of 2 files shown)


fsouza:
Thanks for contributing. Would you be able to add a test that fails with the previous code and passes with the fix. Ideally something using the SDK rather than asserting on the value of the headers.

yuki2006:
@fsouza thx comment.

I don't know how to use the SDK, but I wrote some unit tests.

Please feel free to make any changes.

fakestorage/object.go
```diff
@@ -686,7 +687,7 @@ func (s *Server) handleRange(obj Object, r *http.Request) (ranged bool, start in
 				if end > int64(len(obj.Content)) {
 					end = int64(len(obj.Content))
 				}
-				return true, start, end, obj.Content[start:end]
+				return true, start, end - 1, obj.Content[start:end]
```
StevenACoffman, line 690: Shouldn't this be:
```
return true, start, end - 1, obj.Content[start:end-1]
```
?

fsouza: Nope, in the slice expression the upper bound is non-inclusive: https://go.dev/ref/spec#Slice_expressions

PR #803, opened 2022-06-01
Handle offset-from-end ranges and make range handling more robust

### LINKED ISSUE #536: Add support for negative offset
It seems that negative offset is not considered, such as `range=-10-` which means `NewRangeReader(ctx, -10, -1)`.
https://github.com/fsouza/fake-gcs-server/blob/5a7dc5a5a672ab1b61ed7a9d4180614d60245975/fakestorage/object.go#L578-L599

It can be seen in `Example(LastNBytes)` in https://pkg.go.dev/cloud.google.com/go/storage#ObjectHandle.NewRangeReader

dnerdy:
This PR adds support for offset-from-end ranges. An offset-from-end
range header looks like "Range: bytes=-10". This returns content
starting 10 bytes from the end of the content to the end of the content.

The PR also updates other range responses to align with the real GCS
server. Specifically, the range handling code now detects when the start
of range is beyond the content bounds and returns a 416 response. Most
other invalid ranges are ignored, except for a few cases where bounds
are automatically adjusted.

All of the cases were verified against the GCS server using requests
like:
```
curl -D - https://storage.googleapis.com/marksandstrom-test/test.txt -H"Range: bytes=0-4"
```

Fixes #536.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -683,7 +683,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 
 func (s *Server) handleRange(obj Object, r *http.Request) (ranged bool, start int64, lastByte int64, content []byte, satisfiable bool) {
 	contentLength := int64(len(obj.Content))
-	start, end, err := _parseRange(r.Header.Get("Range"), contentLength)
+	start, end, err := parseRange(r.Header.Get("Range"), contentLength)
 	if err != nil {
 		// If the range isn't valid, GCS returns all content.
 		return false, 0, 0, obj.Content, false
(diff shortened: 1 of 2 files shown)


StevenACoffman:
@fsouza Looks good to me. Anything else you'd like here?

fakestorage/object.go
```diff
@@ -648,50 +647,135 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
...
+	parts := strings.SplitN(rangeHeaderValue, "=", 2)
+	if len(parts) != 2 {
+		return 0, 0, fmt.Errorf("expecting `=` in range header, got: %s", rangeHeaderValue)
+	}
+	if parts[0] != "bytes" {
+		return 0, 0, fmt.Errorf("invalid range unit, expecting `bytes`, got: %s", parts[0])
+	}
+	rangeSpec := parts[1]
+	if len(rangeSpec) == 0 {
+		return 0, 0, fmt.Errorf("empty range")
```
fsouza, line 751: nit: since we're not formatting the string, can use errors.New
```suggestion
		return 0, 0, errors.New("empty range")
```

fakestorage/object.go
```diff
@@ -648,50 +647,135 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
...
+		return true, start, end, obj.Content[start : end+1], true
+	}
+}
+
+// parseRange parses the range header and returns the corresponding start and
+// end indices in the content. The end index is inclusive. This function
+// doesn't validate that the start and end indices fall within the content
+// bounds. The content length is only used to handle "suffix length" and
+// range-to-end ranges.
+func parseRange(rangeHeaderValue string, contentLength int64) (start int64, end int64, err error) {
```
fsouza, line 727: probably worth it adding some tests specifically for this function, specially for the edge cases. I can take care of it later if you prefer though.

fsouza (approved):

fakestorage/object.go
StevenACoffman: @fsouza  If you are adding these tests, the std lib has (unexported unfortunately for our needs): https://github.com/golang/go/blob/085529bd5fa3ab508784f6a3d42f15d2dafddc65/src/net/http/fs.go#L870
which is tested here:
https://github.com/golang/go/blob/085529bd5fa3ab508784f6a3d42f15d2dafddc65/src/net/http/fs_test.go#L102

So you might be able to crib some test cases and save some time.

PR #805, opened 2022-06-02
Return a 409 when attempting to create a bucket that already exists

### LINKED ISSUE #804: Creating a bucket that already exists should return a 409
Both the fs and memory backend return 200 responses when creating a bucket that already exists (the fs backend uses os.MkdirAll and the memory backend allows re-creation of existing buckets as long as VersioningEnabled matches).

GCS returns a 409 response when attempting to create a bucket that already exists. It would be nice if fake-gcs-server matched this behavior.

dnerdy:
This PR updates createBucketByPost to return a 409 response when the
bucket being created already exists. This matches the behavior of the
GCS API.

Fixes #804.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -7,6 +7,7 @@ package fakestorage
 import (
 	"encoding/json"
 	"errors"
+	"fmt"
 	"net/http"
 	"regexp"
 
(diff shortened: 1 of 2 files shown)


fakestorage/bucket.go
```diff
@@ -70,6 +71,18 @@ func (s *Server) createBucketByPost(r *http.Request) jsonResponse {
 		return jsonResponse{errorMessage: err.Error(), status: http.StatusBadRequest}
 	}
 
+	_, err := s.backend.GetBucket(name)
```
fsouza, line 74: I was wondering if we should have the backend return an error or something like that (to avoid TOCTOU). This is an emulator though, I imagine TOCTOU isn't really an issue, and if it becomes an issue we can fix it x)

PR #813, opened 2022-06-05
Use dynamic matcher in signed url uploads

KieranHarper:
So that signed url uploads work if you change the publicHost after the server starts.


fsouza (approved):
Thanks!

PR #823, opened 2022-06-07
return attrs for newly created object in rewrite

jess-sheneberger:
When using [Copier.Run](https://pkg.go.dev/cloud.google.com/go/storage#Copier.Run) to copy a file the resulting attributes object had 0 for generation.  This change seems to populate the generation field with the value from the created object, but I'm not certain it's actually the correct fix--maybe the other attributes should be from newObject?

Please let me know if you need more information about the issue or the change. The tests appear to be passing.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -624,12 +624,12 @@ func (s *Server) rewriteObject(r *http.Request) jsonResponse {
 		Content: append([]byte(nil), obj.Content...),
 	}
 
-	_, err = s.createObject(newObject)
+	created, err := s.createObject(newObject)
 	if err != nil {
 		return errToJsonResponse(err)
 	}
 
-	return jsonResponse{data: newObjectRewriteResponse(newObject.ObjectAttrs)}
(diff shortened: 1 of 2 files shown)


fsouza (commented):
Hmm that should work yeah. Any chance we can add a test?

Thanks for contributing!

jess-sheneberger:
I just pushed a change to the existing rewrite tests to ensure that generation is not zero.  I'm not sure if it's possible to predict in advance the generation. Maybe the correct test is to try fetching the version of the file with the returned generation? Let me know what you think.

I verified that the updated tests fail when I revert my change.

fsouza:
> I just pushed a change to the existing rewrite tests to ensure that generation is not zero. I'm not sure if it's possible to predict in advance the generation. Maybe the correct test is to try fetching the version of the file with the returned generation? Let me know what you think.

This looks good, can you just fix the golangci-lint failure? Thanks for contributing!

fsouza (approved):

fsouza:
Thank you!

Issue #397, opened 2021-01-21
Docker gRPC FUSE + Large file in Docker bind mount = Memory leak 


ChrisWhealy:
I think this issue is more of a gotcha that should be documented rather than a bug in the coding...

I need to use `fake-gcs-server` for serving up some CSV files.  Most of these are pretty small (< 2Mb), but one is >6Gb

If I start the container with `data` bind mounted to the directory containing only the small CSV files, then everything is fine.  However, if I add the very large CSV to the bind mount directory, then at start up, the container's memory usage just climbs until it hits the allocation limit and crashes.

In the Docker desktop, if I switch off the experimental feature "Use gRPC FUSE for file sharing", then the container starts up without this run-away memory usage.

fsouza:
What command are you running to start the container? Are you using the memory backend?

ChrisWhealy:
I'm starting the container with the command

`docker run -d --memory=16g --name fake-gcs-server -p 4443:4443 -v ${PWD}/data:/data fsouza/fake-gcs-server:latest`

The Docker VM is set to use 20Gb and "Use gRPC FuSE for file sharing" is switched off

fsouza:
This is probably not a memory leak, the code is just not very smart when loading stuff from the disk. There were some improvements to the list endpoint recently, but there's definitely room to improve other parts of the code.

warent:
I'm mounting an empty directory, so the bucket has no contents.

<img width="410" alt="Screen Shot 2021-11-11 at 10 01 22 AM" src="https://user-images.githubusercontent.com/13342266/141338784-cc902459-2b80-469c-8a92-b297bae97169.png">

```
command: -port 8080 -public-host gcs:8080 -backend filesystem -filesystem-root /data
volumes:
  - ./buckets:/data
```

<img width="295" alt="Screen Shot 2021-11-11 at 10 04 24 AM" src="https://user-images.githubusercontent.com/13342266/141338983-faad9067-cf70-40bf-aaf6-9da4fe9fca07.png">

I had to force quit because my 16gb macbook pro isn't powerful enough for a single go program running in 1 docker container somehow? This is definitely not normal

warent:
I cloned the repo and built the image locally. No idea why but it seems to be working a bit better now. Memory use is still crazy but not as bad. Above it was using like 11gb, now it's fine with like 5g

I ran a profiler and here's the result.

[...]

PR #829, opened 2022-06-15
Stream object downloads instead of buffering objects in memory

### LINKED ISSUE #828: Efficiently handle large objects when downloading
This issue is like #669 except that it's specific to downloading. When downloading an object fake-gcs-server loads an entire object into memory. This can be problematic when there are concurrent requests of large objects. It's especially problematic when there are concurrent range reads. We have code that performs a bunch of small concurrent range reads on a large file, and responses that are at most a few kilobytes in size cause tens of gigabytes of data to be loaded into memory.

dnerdy:
Fixes #828.

This PR makes it so objects in the FS backend are streamed from disk
instead of being loaded into memory. This drastically reduces memory
usage when working with large files, especially when performing small
concurrent range reads on a large file.

A StreamingObject type has been added to both the fakestorage and
backend packages. This type contains a io.ReadSeekCloser that can be
used to access object data.

This change sets the stage for fixing #669 and #397. #669 isn't fixed
because it will require re-working of how resumable uploads are stored
(though other upload types have been made streaming in this PR). Also,
fixing #397 would require changing how initial objects are specified
when starting the server, which with this approach would require
changing the public API.

Regarding the public API: the server methods for working with objects
have been changed to use StreamingObject instead of Object. This is a
breaking change. It would be possible to make streaming backward
compatible by adding some metadata and methods to Object (to distinguish
between streaming and buffered objects), but I went with this approach
because it made identifying the code that needed to be updated easier;
the compiler pointed out the locations to update instead of needing to
track down runtime and test errors. An additional improvement, which is
compatible with the backward compatible approach, would be to add an
OpenForReading method on Object so callers that are currently
responsible for closing objects also explicitly open objects (which
would make the responsibility clearer). Anyway, I wanted to get this
initial approach up for feedback before putting in additional work.


StevenACoffman:
@fsouza what do you think of this? It has made a very large performance improvement in our internal testing.

dnerdy:
Our engineers have been using this change in our dev environment for about a week (probably millions of requests). Folks use Linux and macOS. It would be helpful if someone in the community could also test the change on Windows. Windows has very different file open, rename and delete semantics. I've used github.com/alexbrainman/goissue34681 (found via https://github.com/golang/go/issues/34681), though I'm not sure if this works well with multiple readers.

The strategy taken in this code is as follows:

- Obtain lock when accessing an object
- Open backing file
- Unlock
- If later the object is updated or deleted, the backing file is replaced or deleted (under a different lock)
- Open readers (which no longer hold a lock) continue to read from the original open backing file

This allows a slow reader (or multiple readers) to download the original file even if the backing file has been modified or deleted.

fsouza:
> @fsouza what do you think of this? It has made a very large performance improvement in our internal testing.

On a quick glance, this looks sane. I'm a bit slow to respond right now, but I'll try to come back to it later today or some time tomorrow.

I also don't have access to any Windows host, so if we don't get any volunteers the best strategy will probably be to merge the PR and wait for bug reports 🙈

fsouza (commented):
Looks good overall, just want to make sure we don't ship a breaking change in the lib API.

fakestorage/upload.go
```diff
@@ -411,7 +404,7 @@ func (s *Server) resumableUpload(bucketName string, r *http.Request) jsonRespons
 
 // uploadFileContent accepts a chunk of a resumable upload
 //
-// A resumable upload is sent in one or more chunks. The request's
+// A resumeable upload is sent in one or more chunks. The request's
```
StevenACoffman, line 407:
```suggestion
// A resumable upload is sent in one or more chunks. The request's
```
Typo? [resumable](https://www.merriam-webster.com/dictionary/resumable) is correct

dnerdy:
@fsouza are you still interested in merging this PR? I'm curious what your opinion is about other breaking changes in this PR. You mentioned CreateObject, and @StevenACoffman changed the name back in #882, though the function signature is still different: it takes a StreamingObject. Some of the other functions also have different signatures. Are you okay with any breaking changes, or do you want the exported functions to stay backward compatible (which would mean creating new streaming variants)?

fsouza:
> @fsouza are you still interested in merging this PR? I'm curious what your opinion is about other breaking changes in this PR. You mentioned CreateObject, and @StevenACoffman changed the name back in #882, though the function signature is still different: it takes a StreamingObject. Some of the other functions also have different signatures. Are you okay with any breaking changes, or do you want the exported functions to stay backward compatible (which would mean creating new streaming variants)?

Hey, apologies for the silence, I was on a long break, but coming back now! Yeah, I wanted to keep it fully backwards compatible. We could keep the streaming variant private unless yall are already using fake-gcs-server as a library for this. Like, keep it private unless someone actually needs it? Not sure what you think.

dnerdy:
> We could keep the streaming variant private unless yall are already using fake-gcs-server as a library for this. Like, keep it private unless someone actually needs it? Not sure what you think.

No worries! And apologies for taking a while to circle back to this myself. Yeah, that seems reasonable. I'll work on reverting the existing APIs and I'll double check our usage to see if anything new needs to be exported (I suspect not).

fsouza:
@dnerdy I'm curious on how you and @StevenACoffman would like to proceed. Should we close #882 and continue working on this PR?

dnerdy:
@fsouza The original buffered APIs have been restored:
- CreateObject
- GetObject
- GetObjectWithGeneration

The new streaming variants are named `<X>Streaming`, i.e.
- CreateObjectStreaming
- GetObjectStreaming
- GetObjectWithGenerationStreaming

Internally, the server uses streaming objects and the objects are converted to buffered objects in the restored methods.

I've moved the documentation from the buffered variants to the streaming variants, to encourage their use, and updated the documentation for the buffered methods to point to the streaming variants.

One additional change I made to CreateObjectStreaming is that it returns an error instead of panicking when an error occurs.

I've also merged in main and updated the code to use the new metadata interface.

internal/backend/fs.go
```diff
@@ -152,26 +162,66 @@ func (s *storageFS) CreateObject(obj Object) (Object, error) {
...
 	if err != nil {
-		return Object{}, err
+		return StreamingObject{}, err
 	}
 
 	path := filepath.Join(s.rootDir, url.PathEscape(obj.BucketName), url.PathEscape(obj.Name))
 
-	if err = os.WriteFile(path, obj.Content, 0o600); err != nil {
-		return Object{}, err
+	tempFile, err := os.CreateTemp("", "fake-gcs-object")
```
fsouza, line 170: If we want to guarantee that rename will work, we'll likely want to specify the rename, otherwise the temporary directory may be in a different mount point. No need to overthink it though, we could address it later if someone raises the issue :)

internal/backend/fs.go
```diff
@@ -152,26 +162,66 @@ func (s *storageFS) CreateObject(obj Object) (Object, error) {
...
+	tempFile, err = compatOpenForWritingAllowingRename(tempFile.Name())
+	if err != nil {
+		return StreamingObject{}, err
+	}
+	defer tempFile.Close()
+
+	// The file is renamed below, which causes this to be a no-op. If the
+	// function returns before the rename, though, the temp file will be
+	// removed.
+	defer os.RemoveAll(tempFile.Name())
```
fsouza, line 185: nit: os.Remove should be enough?
```suggestion
	defer os.Remove(tempFile.Name())
```

internal/backend/rename_unix.go
```diff
@@ -0,0 +1,21 @@
+//go:build !windows
+// +build !windows
```
fsouza, line 2: Hmm now that we only need to support Go 1.18 and 1.19, we can get rid of `+build` (`go fix` should do it automatically I think).

internal/backend/rename_windows.go
```diff
@@ -0,0 +1,54 @@
+package backend
+
+import (
+	"fmt"
+	"os"
+	"path/filepath"
+	"time"
+
+	"github.com/alexbrainman/goissue34681"
```
fsouza, line 9: This feels wrong lol I wonder if the author will keep that around. I guess that in the worst case it'll be cached in the goproxy and we can fork it.

fsouza (commented):
This looks good in general, I have some minor comments left. Thank you very much for pushing this through!

fsouza (approved):

PR #832, opened 2022-06-15
Add rename test to experiment with Windows behavior

dnerdy:
This is a test PR so I debug atomic file rename behavior that's failing on Windows via GitHub actions. Apologies for the noise.


dnerdy:
Okay. I understand the issue now, and using the `goissue34681` seems like a reasonable solution.

PR #833, opened 2022-06-16
Fix typo in readme

muminovicamir:
There was a typo when trying to run the HTTP version locally. I resolved it by adding an additional dash in the proposed section.


fsouza (commented):
`--scheme` should not work. The readme is correct. Can you provide more details on what you tried and what happened?

muminovicamir:
You're right, I had a different typo. Thanks for the quick answer!

PR #843, opened 2022-06-26
Fix upload response structure

barathsriram:
The API response returned for the Upload Endpoints by the fake-gcs-server differs from the response documented in Google for the `insert` endpoint. 

Reference: https://cloud.google.com/storage/docs/json_api/v1/objects/insert#response


yottabytt:
we are facing the same issue. would be great if this gets merged.

yottabytt:
we are facing the same issue when calling these APIs from rust. would be great if this gets merged.

fsouza (approved):
Looks good, thanks for fixing.

Do you think we could add a failing test? Or a failing example to the examples folder? I can merge this one and we can take care of testing in a follow-up though.

Issue #532, opened 2021-07-13
Resumable uploads don't set content-type


fsouza:
Somehow exposed by #530, but resumable uploads don't set content-type and an empty Content-Type header can break the .NET SDK in that specific example.

Additionally, we should consider not setting the content-type header if we don't have it.

aheritier:
I confirm that real uploads on GCS are setting the content-type but somehow it is lost when using fake-gcs-server

PR #851, opened 2022-07-08
Add support for gzip transcoding

ekimekim:
When an object has Content-Encoding: gzip, and an object download request does not have Accept-Encoding: gzip, GCS will decompress the object before serving it. This is documented here: https://cloud.google.com/storage/docs/transcoding

I've added support for this alongside the current range request logic (it's related - if we are transcoding, we ignore the range header).

To make this work for our use-case I also had to fix a bug in resumable uploads where content encoding metadata was not being set if it was specified in the request body.

EDIT: I just noticed https://github.com/fsouza/fake-gcs-server/issues/532, I believe this PR fixes that as well.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -5,6 +5,8 @@
 package fakestorage
 
 import (
+	"compress/gzip"
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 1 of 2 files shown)


fsouza (commented):
Thanks for contributing and apologies for the delayed response. Can you fix the lint job and add a test or two? Thanks!

PR #873, opened 2022-07-29
Do not read the objects data on list

ironsmile:
I am back with my object listing improvements :smile: 

This optimization takes advantage of the fact that the object metadata
could now be read from disk without reading its data. Previously the
whole file was read from disk in order to obtain its size and nothing
more. Instead, now a simple Stat is used.


fsouza:
Hey @ironsmile, thanks for keeping working on this! Can you take a look at the test failure?

ironsmile:
Certainly! Did not realize there are failing tests originally.

fsouza (approved):

PR #882, opened 2022-08-04
Streaming downloads

### LINKED ISSUE #828: Efficiently handle large objects when downloading

StevenACoffman:
Fixes #828. (This is #829 with the review requests applied)

This PR makes it so objects in the FS backend are streamed from disk instead of being loaded into memory. This drastically reduces memory usage when working with large files, especially when performing small concurrent range reads on a large file.

A StreamingObject type has been added to both the fakestorage and backend packages. This type contains a io.ReadSeekCloser that can be used to access object data.

This change sets the stage for fixing #669 and #397. #669 isn't fixed because it will require re-working of how resumable uploads are stored (though other upload types have been made streaming in this PR). Also, fixing #397 would require changing how initial objects are specified when starting the server, which with this approach would require changing the public API.

Regarding the public API: the server methods for working with objects have been changed to use StreamingObject instead of Object. This is a breaking change. It would be possible to make streaming backward compatible by adding some metadata and methods to Object (to distinguish between streaming and buffered objects), but I went with this approach because it made identifying the code that needed to be updated easier; the compiler pointed out the locations to update instead of needing to track down runtime and test errors. An additional improvement, which is compatible with the backward compatible approach, would be to add an OpenForReading method on Object so callers that are currently responsible for closing objects also explicitly open objects (which would make the responsibility clearer). Anyway, I wanted to get this initial approach up for feedback before putting in additional work.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,1 +1,2 @@
 fake-gcs-server
+.idea
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -5,6 +5,7 @@
 package fakestorage
 
 import (
+	"bytes"
 	"encoding/json"
 	"errors"
 	"fmt"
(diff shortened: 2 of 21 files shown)


StevenACoffman:
@dnerdy @fsouza I have applied the requested changes from #829

StevenACoffman:
Closing to continue work in #829

Issue #894, opened 2022-08-18
Sharing volume on macOS: "failed to determine if storage supports xattr"


tdeg:
Hi. I am trying to mount a volume between my host machine and docker container. My goal is be able to see the data that is getting uploaded outputted to a folder in my project's directory. After reading through the codebase my current configuration looks like this:

```
services:
  my_service:
    build:
      context: "."
      dockerfile: dev.Dockerfile
    depends_on:
        - db
    volumes:
      - .:/myservice
      - ./storage:/storage
    environment:
      STORAGE_EMULATOR_HOST: http://storage:4443
    ports:
      - 8080:8080
    networks:
      - network

  storage:
    image: fsouza/fake-gcs-server
    volumes:
      - ./storage:/storage
    ports:
      - 4443:4443
    networks:
      - network
    command:
      - "-scheme" 
      - "http"
      - "-backend" 
      - "filesystem"
  networks:
    network:
```

When running compose up, the container crashes with the following error: 
```
level=fatal msg="couldn't start the server" error="failed to determine if \"/storage/\" supports xattr, consider using a different storage backend or a filesystem that supports it (e.g. not tmpfs on Linux): xattr.get /storage/ user.metadata: operation not supported"
```

If I remove the volume from the storage definition the container builds and I can access the buckets via cURL. Any help is greatly appreciated.

fsouza:
Hmm yeah that won't work with the way we store metadata in the filesystem backend (we use extended attributes, which are [not supported by Docker for Mac](https://docker-docs.netlify.app/docker-for-mac/osxfs/#extended-attributes). What we usually recommend is to use the shared volume as the source and let fake-gcs-server manage the storage root, but in that case you wouldn't be able to see the files/directories in the host. Can you provide more details on your use case so we can figure something out?

tdeg:
Ah okay. For the time being I will verify uploads via cURL but my use case is web scraping so I would love to be able to see the actual data that's being scraped outputted. I have other ways of doing this but was trying to simulate GCP much as possible locally.

fsouza:
We do have a fallback in place for Windows where xattr isn't supported at all (we store the metadata in a JSON file). Perhaps what we could do is fallback automatically to the json format whenever the filesystem root doesn't support xattr, instead of assuming that anything that's not Windows supports it. Shouldn't be too hard I think

fsouza:
@tdeg can you try #897?

tdeg:
[...]

PR #897, opened 2022-08-18
backend/fs: fallback to file-based metadata when xattr isn't available

fsouza:
Related to #894.


internal/backend/metadata_file.go
```diff
@@ -0,0 +1,30 @@
+// Copyright 2022 Francisco Souza. All rights reserved.
```
tdeg, line 1: 👍🏻

internal/backend/metadata_xattr.go
```diff
@@ -0,0 +1,29 @@
+// Copyright 2022 Francisco Souza. All rights reserved.
```
tdeg, line 1: 👍🏻

tdeg (approved):

Issue #349, opened 2020-10-21
Incorrect handling of paths which start with a /


jeantil:
I have updated the sample in #345 to add the ability to log http requests.

According to the [documentation](https://cloud.google.com/storage/docs/naming-objects) object names must adhere to some rules but may start with a / 
While I don't expect this project to fully follow the naming rules in the documentation, incorrect handling of paths which start with a / is problematic because such paths are used as exemples in the documentation (such as `/europe/france/paris.jpg`) which can be a bad experience for begginers. 

below are HTTP traces for both cases 

* writing `europe/france/paris.jpg` :heavy_check_mark: 
```
POST /upload/storage/v1/b/bucket_1603295698243689564/o?uploadType=multipart&name=europe%2Ffrance%2Fparis.jpg
user-agent: gcloud-node-storage/5.3.0
x-goog-api-client: gl-node/12.18.3 gccl/5.3.0
authorization: [...redacted...]
content-type: multipart/related; boundary=1a1f5d37-cb92-4c2a-9d46-5d3b17f45c03
accept: */*
accept-encoding: gzip,deflate
host: 127.0.0.1:4443
--1a1f5d37-cb92-4c2a-9d46-5d3b17f45c03
Content-Type: application/json

{"contentType":"image/jpeg"}
--1a1f5d37-cb92-4c2a-9d46-5d3b17f45c03
Content-Type: image/jpeg

fake-gcs-server is awesome
--1a1f5d37-cb92-4c2a-9d46-5d3b17f45c03--

HTTP/1.1 200
date: Wed, 21 Oct 2020 15:54:58 GMT
content-length: 357
content-type: text/plain; charset=utf-8
{"name":"europe/france/paris.jpg","contentType":"image/jpeg","contentEncoding":"","crc32c":"qHNz6Q==","md5Hash":"nBHYLqMw7b5xFMag+TA1/Q==","acl":[{"Entity":"projectOwner","EntityID":"","Role":"OWNER","Domain":"","Email":"","ProjectTeam":null}],"created":"2020-10-21T15:54:58.30484Z","updated":"2020-10-21T15:54:58.304845Z","deleted":"0001-01-01T00:00:00Z"}
``` 
* reading `europe/france/paris.jpg` :heavy_check_mark: 
```
GET /storage/v1/b/bucket_1603295698243689564/o/europe%2Ffrance%2Fparis.jpg?alt=media
accept-encoding: gzip
cache-control: no-store
[...]

PR #918, opened 2022-09-12
Fix support resumable + fix incorrect handling of paths which start with a / 

le0pard:
Should fix support for `resumable` + https://github.com/fsouza/fake-gcs-server/issues/349

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -213,7 +213,7 @@ func newServer(options Options) (*Server, error) {
 
 func (s *Server) buildMuxer() {
 	const apiPrefix = "/storage/v1"
-	s.mux = mux.NewRouter()
+	s.mux = mux.NewRouter().SkipClean(true).UseEncodedPath()
 
 	// healthcheck
 	s.mux.Path("/_internal/healthcheck").Methods(http.MethodGet).HandlerFunc(s.healthcheck)
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -251,8 +251,8 @@ func (s *Server) buildMuxer() {
 	s.mux.Host(bucketHost).Path("/{objectName:.+}").Methods(http.MethodGet, http.MethodHead).HandlerFunc(s.downloadObject)
 	s.mux.Path("/download/storage/v1/b/{bucketName}/o/{objectName:.+}").Methods(http.MethodGet).HandlerFunc(s.downloadObject)
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods(http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.insertObject))
-	s.mux.Path("/upload/resumable/{uploadId}").Methods(http.MethodPut, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.uploadFileContent))
 	s.mux.Path("/upload/storage/v1/b/{bucketName}/o").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.uploadFileContent))
+	s.mux.Path("/upload/resumable/{uploadId}").Methods(http.MethodPut, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.uploadFileContent))
 
 	// Batch endpoint
 	s.mux.MatcherFunc(s.publicHostMatcher).Path("/batch/storage/v1").Methods(http.MethodPost).HandlerFunc(s.handleBatchCall)
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -19,6 +19,7 @@ import (
 	"net/http/httptest"
 	"net/http/httputil"
 	"net/textproto"
+	"net/url"
 	"strings"
 	"sync"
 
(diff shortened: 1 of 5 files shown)


fakestorage/upload.go
```diff
@@ -203,9 +203,19 @@ func (s *Server) checkUploadPreconditions(r *http.Request, bucketName string, ob
...
 }
 
+func normalizeFileName(name string) string {
+	for {
+		name = strings.Replace(name, "/", "", 1)
+		if !strings.HasPrefix(name, "/") {
+			break
+		}
+	}
+	return name
```
fsouza, line 213: Could use strings.TrimLeft. I don't think this is what we want though. Like, doesn't GCS treat `file.txt` and `/file.txt` as different objects?

le0pard: @fsouza it is the same objects

le0pard: Changed to `strings.TrimLeft`

le0pard: Some libraries even remove leading slashes from own side - https://github.com/googleapis/google-cloud-node/pull/2269/files

le0pard: Fixed in different way - https://github.com/fsouza/fake-gcs-server/pull/918/commits/41e8db7594af4433bf6b0818ef463df0764cd16d

le0pard:
Now tests should pass

fsouza:
Hey thanks for contributing. I'm going to sit down and review this this weekend.

fakestorage/server.go
```diff
@@ -211,9 +212,22 @@ func newServer(options Options) (*Server, error) {
...
+			m[k] = v
+		}
+	}
+	return m
+}
+
 func (s *Server) buildMuxer() {
 	const apiPrefix = "/storage/v1"
-	s.mux = mux.NewRouter()
+	s.mux = mux.NewRouter().SkipClean(true).UseEncodedPath()
```
le0pard, line 230: Reason for this.

> UseEncodedPath tells the router to match the encoded original path to the routes. For eg. "/path/foo%2Fbar/to" will match the path "/path/{var}/to".

This fix issue, that if we use names like `/file.txt`, in this case it will be `/download/storage/v1/b/{bucketName}/o/%2Ffile.txt` in url path, which mux trying to fix by redirect to `/download/storage/v1/b/{bucketName}/o/file.txt`, which is incorrect behaviour for GCP

fakestorage/server.go
```diff
@@ -211,9 +212,22 @@ func newServer(options Options) (*Server, error) {
 	return &s, nil
 }
 
+func unescapeMuxVars(vars map[string]string) map[string]string {
```
le0pard, line 215: In some places we use `r.URL.Query()`, which provide already decoded values. But `mux.Vars(r)` provide values "as is" (without decoding). That is why we need decode `objectName` and `bucketName` ourself. This is important, because before accessing to storage we encode values and this lead to double encoding.

fakestorage/upload.go
```diff
@@ -391,9 +395,16 @@ func (s *Server) resumableUpload(bucketName string, r *http.Request) jsonRespons
 	}
 	s.uploads.Store(uploadID, obj)
 	header := make(http.Header)
-	header.Set("Location", s.URL()+"/upload/resumable/"+uploadID)
+	location := fmt.Sprintf(
+		"%s/upload/storage/v1/b/%s/o?uploadType=resumable&name=%s&upload_id=%s",
+		s.URL(),
+		url.PathEscape(bucketName),
```
le0pard, line 401: we have to encode values for url

fakestorage/upload.go
```diff
@@ -362,6 +363,9 @@ func (s *Server) multipartUpload(bucketName string, r *http.Request) jsonRespons
 }
 
 func (s *Server) resumableUpload(bucketName string, r *http.Request) jsonResponse {
+	if r.URL.Query().Has("upload_id") {
+		return s.uploadFileContent(r)
```
le0pard, line 367: if `upload_id` came - it mean we have request to upload file.

le0pard:
Thanks @fsouza - added comments to simplify review

fsouza (commented):
Thanks for contributing! The change looks sane, but can you add some tests? Ideally something that would fail before this change, but passes with both storage backends!

le0pard:
@fsouza added test cases which fail in current master:

```bash
--- FAIL: TestServerClientObjectWriter (0.01s)
    --- FAIL: TestServerClientObjectWriter/https_listener (0.40s)
        --- FAIL: TestServerClientObjectWriter/https_listener/file_with_backslash_at_beginning (0.04s)
            upload_test.go:120: storage: object doesn't exist
        --- FAIL: TestServerClientObjectWriter/https_listener/file_with_backslashes_at_name (0.13s)
            upload_test.go:120: storage: object doesn't exist
    --- FAIL: TestServerClientObjectWriter/no_listener (0.40s)
        --- FAIL: TestServerClientObjectWriter/no_listener/file_with_backslash_at_beginning (0.05s)
            upload_test.go:120: storage: object doesn't exist
        --- FAIL: TestServerClientObjectWriter/no_listener/file_with_backslashes_at_name (0.12s)
            upload_test.go:120: storage: object doesn't exist
    --- FAIL: TestServerClientObjectWriter/http_listener (0.40s)
        --- FAIL: TestServerClientObjectWriter/http_listener/file_with_backslash_at_beginning (0.05s)
            upload_test.go:120: storage: object doesn't exist
        --- FAIL: TestServerClientObjectWriter/http_listener/file_with_backslashes_at_name (0.12s)
            upload_test.go:120: storage: object doesn't exist
```

fakestorage/upload.go
fsouza, line 401: Nit: I assume that the bucket doesn't really need to be url-encoded? See https://cloud.google.com/storage/docs/naming-buckets
```suggestion
		bucketName,
```

fsouza (approved):
One last comment and we should be good to go. Thank you very much!

fsouza (approved):
Thank you!

PR #935, opened 2022-09-29
Race-free Preconditions

rawler:
This fixes an issue we've experienced, where "IfNotExist"-style uploads have exhibited racy behavior, [specified to not exist in production GCS](https://cloud.google.com/storage/docs/consistency).

### DIFF
--- a/fakestorage/json_response.go
+++ b/fakestorage/json_response.go
@@ -6,6 +6,8 @@ import (
 	"net/http"
 	"os"
 	"syscall"
+
+	"github.com/fsouza/fake-gcs-server/internal/backend"
 )
 
 type jsonResponse struct {
@@ -62,6 +64,10 @@ func errToJsonResponse(err error) jsonResponse {
 	var pathError *os.PathError
 	if errors.As(err, &pathError) && pathError.Err == syscall.ENAMETOOLONG {
 		status = http.StatusBadRequest
+	}
+	var backendError backend.Error
+	if errors.As(err, &backendError) && backendError == backend.PreConditionFailed {
+		status = http.StatusPreconditionFailed
 	}
 	return jsonResponse{errorMessage: err.Error(), status: status}
 }
--- a/fakestorage/json_response.go
+++ b/fakestorage/json_response.go
@@ -65,8 +65,7 @@ func errToJsonResponse(err error) jsonResponse {
 	if errors.As(err, &pathError) && pathError.Err == syscall.ENAMETOOLONG {
 		status = http.StatusBadRequest
 	}
-	var backendError backend.Error
-	if errors.As(err, &backendError) && backendError == backend.PreConditionFailed {
+	if err == backend.PreConditionFailed {
 		status = http.StatusPreconditionFailed
 	}
 	return jsonResponse{errorMessage: err.Error(), status: status}
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -160,8 +160,8 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
 	return xmlResponse{status: http.StatusNoContent}
 }
 
-func (s *Server) wrapUploadPreconditions(r *http.Request, bucketName string, objectName string) (*generationCondition, error) {
-	result := &generationCondition{
+func (s *Server) wrapUploadPreconditions(r *http.Request, bucketName string, objectName string) (generationCondition, error) {
+	result := generationCondition{
 		ifGenerationMatch:    nil,
 		ifGenerationNotMatch: nil,
 	}
@@ -170,7 +170,7 @@ func (s *Server) wrapUploadPreconditions(r *http.Request, bucketName string, obj
 	if ifGenerationMatch != "" {
 		gen, err := strconv.ParseInt(ifGenerationMatch, 10, 64)
 		if err != nil {
-			return nil, err
+			return generationCondition{}, err
 		}
 		result.ifGenerationMatch = &gen
 	}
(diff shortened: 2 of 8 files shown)


rawler:
Apologies in advance for my lack of Go experience. I'm sure much of this patch is non-idiomatic, and it is submitted as much as a basis for discussion, as an actual proposal.

fsouza:
Thank you for contributing! I'll take a look at this change over the next couple of days, but there's definitely no reason to apologize, if anything in the code needs improvements, we'll work on it!

fsouza:
@rawler can you take a look at the build failure?

rawler:
> @rawler can you take a look at the build failure?

I looked into them, but I don't understand how it's caused by what I did. From what I see, the cause seems to be:

`bucket_test.go:278: rename C:\Users\RUNNER~1\AppData\Local\Temp\fakestorage-test-root-1941404170\some-bucket\fake-gcs-object3621328271 C:\Users\RUNNER~1\AppData\Local\Temp\fakestorage-test-root-1941404170\some-bucket\img%2Fhi-res%2Fparty-01.jpg: Access is denied.
`

fsouza:
@rawler ugh ok, I'll dig into that. Thank you for your patience!

fsouza:
Hey, sorry, things have been a bit busy on my side, but I believe I'll have a chance to look into it (both the test failure and review the code) this weekend. 

Thank you very much for your contribution and your patience!

fakestorage/json_response.go
```diff
@@ -63,5 +65,9 @@ func errToJsonResponse(err error) jsonResponse {
 	if errors.As(err, &pathError) && pathError.Err == syscall.ENAMETOOLONG {
 		status = http.StatusBadRequest
 	}
+	var backendError backend.Error
+	if errors.As(err, &backendError) && backendError == backend.PreConditionFailed {
```
fsouza, line 69: we can simplify this check:
```suggestion
	if err == backend.PreConditionFailed {
```

internal/backend/memory.go
```diff
@@ -113,6 +113,18 @@ func findObject(obj Object, objectList []Object, matchGeneration bool) int {
 	return -1
 }
 
+// findObject looks for an object in the given list and return the index where it
```
fsouza, line 116: nit: make the comment match the method name
```suggestion
// findLastObjectGeneration looks for an object in the given list and return the index where it
```

fakestorage/upload.go
```diff
@@ -136,72 +152,40 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
...
 	if err != nil {
 		return xmlResponse{errorMessage: err.Error()}
 	}
 	defer obj.Close()
 	return xmlResponse{status: http.StatusNoContent}
 }
 
-func (s *Server) checkUploadPreconditions(r *http.Request, bucketName string, objectName string) *jsonResponse {
+func (s *Server) wrapUploadPreconditions(r *http.Request, bucketName string, objectName string) (*generationCondition, error) {
+	result := &generationCondition{
```
fsouza, line 164: I think we can stop using a pointer here to avoid the dereference later. This is guaranteed to basically never be nil, so no need to use a pointer.
```suggestion
func (s *Server) wrapUploadPreconditions(r *http.Request, bucketName string, objectName string) (generationCondition, error) {
	result := generationCondition{
```

Then we can always return result when there's an error too.

fakestorage/upload.go
```diff
@@ -354,7 +342,8 @@ func (s *Server) multipartUpload(bucketName string, r *http.Request) jsonRespons
 		},
 		Content: notImplementedSeeker{io.NopCloser(io.MultiReader(partReaders...))},
 	}
-	obj, err = s.createObject(obj)
+
+	obj, err = s.createObject(obj, *conditions)
```
fsouza, line 346: with the change above:

```suggestion
	obj, err = s.createObject(obj, conditions)
```

fakestorage/upload_test.go
```diff
@@ -389,6 +389,50 @@ func TestServerClientObjectOperationsFailureToWriteExistingObject(t *testing.T)
 	})
 }
 
+func TestServerClientUploadRacesAreOnlyWonByOne(t *testing.T) {
+	const (
```
fsouza, line 393: Can you use `runServersTest` to make sure we're testing both backends?

fsouza (commented):
Code looks sane. I'm not sure if things will work in the filesystem backend. Also I feel like at some point we should refactor generations out of the backends, but not a blocker for this PR.

I'm trying to figure out the issue with the CI on Windows, but it's quite cryptic 🤔

fakestorage/upload_test.go
rawler: Sure! I worried a little bit about test-performance, given that this test is trying to provoke a race, therefor running multiple iterations. However, I think perhaps we could tune down `repetitions` slightly to compensate. Attaching test-durations from my machine below.

Also pushing a small improvement in test-diagnostics. On conflict, the test will now describe the conflict in more detail.

ok  	github.com/fsouza/fake-gcs-server	0.016s
ok  	github.com/fsouza/fake-gcs-server/fakestorage	1.176s
ok  	github.com/fsouza/fake-gcs-server/internal/backend	0.031s
ok  	github.com/fsouza/fake-gcs-server/internal/checksum	0.010s
ok  	github.com/fsouza/fake-gcs-server/internal/config	0.010s
ok  	github.com/fsouza/fake-gcs-server/internal/notification	0.026s

fsouza (approved):
Thanks for contributing. I'll simply merge this and drop support for Windows. It's simply not possible to safely rename files on Windows, and I don't expect a lot of people to be using fake-gcs-server on Windows (they can always use WSL or Docker).

PR #972, opened 2022-10-28
Format "generation" field in Pubsub notification as a string

danmcgee-soda:
As documented in the storage objects JSON object representation, all fields are "string formatted as the specified value type", including numbers.

https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations

`strconv.FormatInt` was used for the "size" field, but this wasn't done for generation.

We noticed this when doing local testing, and the messages failed JSON unmarshalling from fake-gcs-server, but were properly unmarshalling from real Google Cloud Storage pubsub messages.

Screen Shot 2022-10-28 at 12 20 35 PM

Screen Shot 2022-10-28 at 12 20 39 PM

### DIFF
--- a/internal/notification/event.go
+++ b/internal/notification/event.go
@@ -155,21 +155,21 @@ func (m *PubsubEventManager) publish(o *backend.StreamingObject, eventType Event
 	return nil
 }
 
-// gcsEvent is the payload of a GCS event.  The description of the full object
-// can be found here:
+// gcsEvent is the payload of a GCS event. Note that all properties are string-quoted.
+// The description of the full object can be found here:
 // https://cloud.google.com/storage/docs/json_api/v1/objects#resource-representations.
 type gcsEvent struct {
 	Kind            string            `json:"kind"`
 	ID              string            `json:"id"`
 	Name            string            `json:"name"`
 	Bucket          string            `json:"bucket"`
-	Generation      int64             `json:"generation,omitempty"`
+	Generation      int64             `json:"generation,string,omitempty"`
 	ContentType     string            `json:"contentType"`
 	ContentEncoding string            `json:"contentEncoding,omitempty"`
 	Created         string            `json:"timeCreated,omitempty"`
 	Updated         string            `json:"updated,omitempty"`
 	StorageClass    string            `json:"storageClass"`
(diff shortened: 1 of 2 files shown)


internal/notification/event.go
```diff
@@ -163,7 +163,7 @@ type gcsEvent struct {
 	ID              string            `json:"id"`
 	Name            string            `json:"name"`
 	Bucket          string            `json:"bucket"`
-	Generation      int64             `json:"generation,omitempty"`
+	Generation      string            `json:"generation,omitempty"`
```
fsouza, line 166: Go can do this for you:

```suggestion
	Generation      int64            `json:"generation,string,omitempty"`
```

internal/notification/event.go
```diff
@@ -181,7 +181,7 @@ func generateEvent(o *backend.StreamingObject, eventType EventType, eventTime st
 		ID:              o.ID(),
 		Name:            o.Name,
 		Bucket:          o.BucketName,
-		Generation:      o.Generation,
+		Generation:      strconv.FormatInt(o.Generation, 10),
```
fsouza, line 184:
```suggestion
		Generation:      o.Generation,
```

fsouza (commented):
Thanks for contributing!

danmcgee-soda:
Good callout that JSON marshalling can do it automatically.

I ultimately didn't go that direction, because it opens a real bucket of worms and left things inconsistent with the other fields. If we do that for `generation`, we should also definitely do it for `size`. And that opens up the typing issue of `created` and `updated` and `eventTime` being passed as `string`, rather than `time.Time`, which means other methods then need to be refactored and updated.

All that said- I don't have a huge preference on how it is fixed, as long as ultimately it is emitted as a string and not a number! I just wanted to keep my changeset to the minimal possible one.

danmcgee-soda:
> Let's make it idiomatic with struct tags, I can take care of the inconsistencies in a separate change.

Sounds great- I'll push some revisions later today and get `size` fixed up too!

fsouza:
@danmcgee-soda thank you!

fsouza (approved):
Thank you!

PR #981, opened 2022-11-04
update java readme

Leooo:
- [x] update java readme

(only port 4443 is exposed in Dockerfile)


fsouza (approved):
Thanks for contributing!

PR #1017, opened 2022-12-23
Do not escape the object name in the fs backend

gaul:
Now fake-gcs-server implicitly creates parent directories when creating an object `/foo/bar/baz`.  This allows creating paths longer than `NAME_MAX` (256) and interoperating with existing filesystems.  This approach does not allow some kinds of objects to exist, e.g., `/foo/` with a non-zero object size, although these should be rare.

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -231,7 +231,8 @@ func (s *storageFS) ListObjects(bucketName string, prefix string, versions bool)
 	// TODO: WalkDir more efficient?
 	bucketPath := filepath.Join(s.rootDir, url.PathEscape(bucketName))
 	if err := filepath.Walk(bucketPath, func(path string, info fs.FileInfo, _ error) error {
-		objName := filepath.Rel(bucketPath, path)
+		// Rel() should never return error since path always descend from bucketPath
+		objName, _ := filepath.Rel(bucketPath, path)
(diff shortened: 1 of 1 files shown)


gaul:
Sorry had some trouble with Windows.

internal/backend/fs.go
```diff
@@ -217,28 +227,30 @@ func (s *storageFS) ListObjects(bucketName string, prefix string, versions bool)
...
-	infos, err := os.ReadDir(filepath.Join(s.rootDir, url.PathEscape(bucketName)))
-	if err != nil {
-		return nil, err
-	}
 	objects := []ObjectAttrs{}
-	for _, info := range infos {
+	// TODO: WalkDir more efficient?
+	bucketPath := filepath.Join(s.rootDir, url.PathEscape(bucketName))
+	if err := filepath.Walk(bucketPath, func(path string, info fs.FileInfo, _ error) error {
+		objName := strings.TrimPrefix(path, bucketPath+string(filepath.Separator))
```
fsouza, line 234: Use `filepath.Rel` instead?
```suggestion
		objName := filepath.Rel(bucketPath, path)
```

fsouza (commented):
Thanks for contributing and apologies for the delayed response!

fsouza (approved):

PR #1020, opened 2022-12-25
Custom metadata missing in Public and Signed URLs #1019

### LINKED ISSUE #1019: Custom metadata missing in Public and Signed URLs
I noticed that the existing implementation does not provide the custom metadata with the `x-goog-meta-` prefix as indicated [here](https://cloud.google.com/storage/docs/xml-api/reference-headers#xgoogmeta). 

I am creating this issue with the intention of submitting a PR. However, I am not a golang dev so feel free to turn it down.

uddhav:
Fixes #1019 

NB: I am not a golang dev - feel free to suggest a proper alternative to resolve #1019.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -809,6 +809,9 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", strconv.FormatInt(contentLength, 10))
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
 	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s,md5=%s", obj.Crc32c, obj.Md5Hash))
+	for name, element := range obj.Metadata {
+		w.Header().Set("X-Goog-Meta-"+name, element)
+	}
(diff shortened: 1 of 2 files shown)


fakestorage/object.go
```diff
@@ -809,6 +809,9 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("Content-Length", strconv.FormatInt(contentLength, 10))
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
 	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s,md5=%s", obj.Crc32c, obj.Md5Hash))
+	for name, element := range obj.Metadata {
+		w.Header().Set("X-Goog-Meta-"+name, element)
+	}
 	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
```
fsouza, line 815: nit: let's use `value`for the variable name and move the for loop to the end, just to make the code easier to read
```suggestion
	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
	for name, value := range obj.Metadata {
		w.Header().Set("X-Goog-Meta-"+name, value)
	}
```

fsouza (approved):
Thanks for contributing!

PR #1022, opened 2022-12-27
Make resumable uploads work from browser with signed url

jasonford:
These headers are necessary to make browser originated signed url requests work.

I don't know go, and these edits likely should be handled at a different level/more comprehensively. Happy to modify and repush if you've got feedback. If not, this PR could be a good sketch for better updates to support browser based signed url requests.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -813,6 +813,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	for name, value := range obj.Metadata {
 		w.Header().Set("X-Goog-Meta-"+name, value)
 	}
+	w.Header().Set("Access-Control-Allow-Origin", "*")
 
 	if ranged && !satisfiable {
 		status = http.StatusRequestedRangeNotSatisfiable
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -125,7 +125,7 @@ func NewServerWithOptions(options Options) (*Server, error) {
 		return nil, err
 	}
 
-	allowedHeaders := []string{"Content-Type", "Content-Encoding", "Range"}
+	allowedHeaders := []string{"Content-Type", "Content-Encoding", "Range", "Content-Range"}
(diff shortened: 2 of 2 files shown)


fakestorage/object.go
```diff
@@ -810,6 +810,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
 	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s,md5=%s", obj.Crc32c, obj.Md5Hash))
 	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
+	w.Header().Set("Access-Control-Allow-Origin", "*")
```
fsouza, line 813: Looking at the docs for GCS API, it seems that this should be configured in the bucket and honored by fake-gcs-server? https://cloud.google.com/storage/docs/configuring-cors#rest-apis

jasonford: Thanks for the pointer (and all your excellent work with this project)!

On quick look it seems the necessary work is:
1) add a handler for PATCH to ```/storage/v1/b/BUCKET_NAME?fields=cors``` that stores the configuration somewhere (do you have a pointer to existing code that shows the preferred way to store this type of config?)
2) load that config at this step and set the header.

Seems pretty straightforward, but being new to go and not having too much time to dive deeply, I'd appreciate a pointer to code with the preferred config set/get approach.

fsouza: Ugh, thinking a bit more about this, this one is tricky for the same versioning is tricky with the 

The filesystem backend currently doesn't support storing any bucket attributes, so that would be a requirement. I don't want to make this PR too large for you.

I opened #1025, we can merge this PR as is to get you unblocked.

I've merged some other changes though, so please rebase your PR so we can get it ready to ship.

fsouza (approved):

Issue #932, opened 2022-09-25
Ability to change log level


Meemaw:
By default, `fake-gcs-server` is quite spammy, doing info logging for every request.

Add an ability to change the log level to e.g `warn/error`

fsouza:
That sounds good to me. We can add a new flag (and eventually an environment variable when we solve #561).

RachitSharma2001:
@fsouza @Meemaw I was wondering if I could pick this task up. My idea is to include a `log-level` flag which can be one of three options:
1. all
2. only-warnings-and-errors
3. only-errors
Let me know what you think of this idea and if there is anything else I should change or add.

fsouza:
Hey @RachitSharma2001, I was thinking the flag would be a string and take anything that logrus accept. You could call [`ParseLevel`](https://pkg.go.dev/github.com/sirupsen/logrus#ParseLevel) and if that fails, you report an error to the user indicating invalid log level or something like that.

RachitSharma2001:
That sounds good! I have made a PR for this. Let me know if there is anything that should be changed or added.

PR #1023, opened 2022-12-28
Add ability to change log level

RachitSharma2001:
Addresses #932 

Adds the `-log-level` flag. From my analysis, it seems like there really are only three kinds of logs made by the gcs server:
1. Info logs: Http request logs and some other information logs made by `main.go`
2. Warning logs
3. Error logs

Nonetheless, the user can specify, with a string, one of the 7 accepted levels from the [logrus](https://pkg.go.dev/github.com/sirupsen/logrus) package: fatal, panic, error, warning, info, debug, or trace, and the appropriate log level will be set.

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -40,6 +40,7 @@ type Config struct {
 	bucketLocation      string
 	certificateLocation string
 	privateKeyLocation  string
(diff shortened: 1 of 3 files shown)


fsouza (approved):
Thanks for contributing!

Issue #410, opened 2021-02-04
Setting blob.custom_time doesn't work, returns always None


oittaa:
I have a piece of code like the following:
```
self.client = storage.Client()
self.bucket = self.client.get_bucket(bucket_name)

blob = self.bucket.blob(blob_name)
blob.upload_from_string(value, content_type=content_type)
blob.reload()
blob.custom_time = datetime.datetime.now(datetime.timezone.utc)
blob.patch()
```
If I set `custom_time` before uploading, even the real Google server erases it from the blob. That's why I patch it after the upload. The fake server unfortunately loses it even after `patch`.

```
blob = self.bucket.get_blob(blob_name)
print(blob.custom_time)
```
results in `None`

oittaa:
To me it looks like the `Custom-Time` is completely missing from objects. https://github.com/fsouza/fake-gcs-server/blob/297b8cdcd2f8fac8652f8b20c023ececb1c368b1/fakestorage/object.go#L25-L43

https://cloud.google.com/storage/docs/metadata#custom-time
https://pkg.go.dev/cloud.google.com/go/storage#ObjectAttrsToUpdate

oittaa:
The original Google Cloud issue about missing `custom_time` before patching is tracked here. https://github.com/googleapis/python-storage/issues/372

fsouza:
Thanks for reporting and linking the docs!

mar-pan:
Hello @fsouza is there any chance that this could be assigned to someone and resolved?

fsouza:
I opened #1024 (big round-number!) as a draft, in case someone wants to work on it. It's unlikely I'll be able to pick it back up in the next couple of weeks.

RachitSharma2001:
@fsouza I believe I have a solution that fixes this issue and updates `CustomTime` correctly. I am ready to submit a pr but since my solution is based off the changes I made to [this](https://github.com/fsouza/fake-gcs-server/pull/1016) PR (specifically the changes added to `PatchObject` method in `object.go`) I just want to make sure that those changes are good before submitting a pr to fix this issue.

fsouza:
@RachitSharma2001 hey, I'm going to look at your PR later today (in 2 hours or so). If you prefer, you can open another PR on top of that one, or we can wait.

Apologies for the delayed review.

PR #1036, opened 2023-01-09
Fix custom-time error

RachitSharma2001:
Fixes https://github.com/fsouza/fake-gcs-server/issues/410.

If I now run the following python script:

```
self.client = storage.Client()
self.bucket = self.client.get_bucket(bucket_name)

blob = self.bucket.blob(blob_name)
blob.upload_from_string(value, content_type=content_type)
blob.reload()
blob.custom_time = datetime.datetime.now(datetime.timezone.utc)
blob.patch()
blob = self.bucket.get_blob(blob_name)
print(blob.custom_time)
```

It now does not print `None` but rather prints the correct time.

Let me know if there is anything I should change or add. One thing I was unsure of is that in `fakestorage/object.go`, I did not make any changes to `UpdateObject`, so it still only updates metadata rather than any other attributes like `CustomTime` or `Acl` (unlike the `PatchObject` method). Let me know if `UpdateObject` should be similarly changed.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -42,6 +42,7 @@ type ObjectAttrs struct {
 	Created    time.Time
 	Updated    time.Time
 	Deleted    time.Time
+	CustomTime time.Time
 	Generation int64
 	Metadata   map[string]string
 }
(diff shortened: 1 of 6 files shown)


fsouza (approved):
Nice, thank you very much!

I think it'd be good to update `UpdateObject` too, but we can do that in a follow-up PR. Do you want to do it or should I work on it?

RachitSharma2001:
Thank you! Yeah I can definitely update the `UpdateObject` method. I will make a PR soon.

PR #1041, opened 2023-01-11
A quick fix for enviroment variables config support when running in a container

### LINKED ISSUE #561: Ability to configure server using environment variables
When running this service inside a docker container it would be handy to be able to configure it using env vars instead of cli parameters.

mikechernev:
Hey, 

This partially fixes #561 so that the project can be used in GitHub actions. I will look into using the env vars as a build in thing, because now there's an overlap of default values.

### DIFF
--- a/run.sh
+++ b/run.sh
@@ -1,0 +1,20 @@
+#!/bin/sh
+
+/bin/fake-gcs-server \
+  -backend "${BACKEND:-filesystem}" \
+  -cert-location "${CERT_LOCATION}" \
+  -cors-headers "${CORS_HEADERS}" \
+  -data "${DATA:-/data}" \
+  -event.list "${EVENT_LIST:-finalize}" \
+  -event.object-prefix "${EVENT_OBJECT_PREFIX}" \
+  -event.pubsub-project-id "${EVENT_PUBSUB_PROJECT_ID}" \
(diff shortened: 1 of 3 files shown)


run.sh
```diff
@@ -0,0 +1,20 @@
+#!/bin/sh
+
```
fsouza, line 2: We can prevent a breaking change by checking whether this script was invoked with any parameters. If so, just pass those parameters through. Something like:

```suggestion

if [ -n "$1" ]; then
    exec /bin/fake-gcs-server "$@"
else
    # the command below
fi
```

fsouza (commented):
Maybe prefix the environment variables with something like `FAKE_GCS_`?

run.sh
```diff
@@ -0,0 +1,20 @@
...
+/bin/fake-gcs-server \
+  -backend "${BACKEND:-filesystem}" \
+  -cert-location "${CERT_LOCATION}" \
+  -cors-headers "${CORS_HEADERS}" \
```
fsouza, line 6: Ugh those won't really work when those environment variables aren't set.

I'll close this PR for now, but if anyone is willing to pick it back up, I think it would be better to have environment variables handled directly in the code, to avoid having a shell script with too many conditionals.

PR #1042, opened 2023-01-11
supportsuccess_action_status form field for HTML form uploads

vincentezw:
When uploading an object with a HTML form, the GCS XML api allows us to set a `success_action_status`: [Upload an object with a HTML form](https://cloud.google.com/storage/docs/xml-api/post-object-forms#:~:text=No-,success_action_status,-The%20status%20code)

> The status code that you want Cloud Storage to respond with when an upload is successful. The default is 204, but you can change this to 200 or 201

Presently `fake-gcs-server` ignores this and responds with the default status code 204 (https://github.com/fsouza/fake-gcs-server/blob/main/fakestorage/upload.go#L162)

This PR allows us to specify a preferred return value of 200, 201 or 204.

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -123,6 +123,17 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
 	if contentTypes, ok := r.MultipartForm.Value["Content-Type"]; ok {
 		contentType = contentTypes[0]
 	}
+	var successActionStatus int = http.StatusNoContent
+	if successActionStatuses, ok := r.MultipartForm.Value["success_action_status"]; ok {
+		successInt, err := strconv.Atoi(successActionStatuses[0])
+		if err != nil {
+			return xmlResponse{errorMessage: err.Error(), status: http.StatusBadRequest}
+		}
+		if successInt != http.StatusOK && successInt != http.StatusCreated && successInt != http.StatusNoContent {
+			return xmlResponse{errorMessage: "invalid success action status", status: http.StatusBadRequest}
+		}
+		successActionStatus = successInt
+	}
 	metaData := make(map[string]string)
 	for key := range r.MultipartForm.Value {
 		lowerKey := strings.ToLower(key)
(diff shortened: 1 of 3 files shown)


fakestorage/upload.go
```diff
@@ -123,6 +123,17 @@ func (s *Server) insertFormObject(r *http.Request) xmlResponse {
 	if contentTypes, ok := r.MultipartForm.Value["Content-Type"]; ok {
 		contentType = contentTypes[0]
 	}
+	var successActionStatus int = http.StatusNoContent
```
fsouza, line 126:
```suggestion
	successActionStatus := http.StatusNoContent
```

fakestorage/upload_test.go
```diff
@@ -922,6 +923,7 @@ func TestFormDataUpload(t *testing.T) {
 	var buf bytes.Buffer
 	const content = "some weird content"
 	const contentType = "text/plain"
+	var successActionStatus int = http.StatusNoContent
```
fsouza, line 926:
```suggestion
	successActionStatus := http.StatusNoContent
```

fsouza (commented):
Thanks for contributing!

Looking at the docs, it looks like when the user passes a `201` they should get the object in the response body. Should we do that?

> The status code that you want Cloud Storage to respond with when an upload is successful. The default is 204, but you can change this to 200 or 201. If you choose 200 or 204, Cloud Storage returns an empty document with those status codes. If you choose 201, Cloud Storage returns an XML document with the elements that are described in Response Body Elements. Note: The Adobe Flash player might not handle responses with an empty document body. You should use status code 201 if this is the case.

(from https://cloud.google.com/storage/docs/xml-api/post-object-v2)

vincentezw:
Thanks for the quick 👀  on this. You're absolutely right, we should return an XML document on code 201 (we don't use that so I had overlooked it). I've added this in now, and in my testing the output matches what GCS returns for this.

fakestorage/xml_response.go
```diff
@@ -12,6 +12,14 @@ type xmlResponse struct {
 	errorMessage string
 }
 
+type xmlResponseBody struct {
+	XMLName  xml.Name `xml:"PostResponse"`
+	Bucket   string
+	Etag     interface{}
```
fsouza, line 18: Any reason to use interface{} here instead of the struct declared below?

vincentezw: Not particularly. I get confused with how Go handles nested structs and their assignability, but have adjusted this now. Let me know if you have any comments or a better way of doing it :).

fsouza (approved):

PR #1049, opened 2023-01-13
Update UpdateObject function to handle more attributes

RachitSharma2001:
Builds off #1036

I noticed that there are currently no tests for `UpdateObject`. I would be open to adding one or a few if they are needed.

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -336,16 +336,17 @@ func (s *storageFS) PatchObject(bucketName, objectName string, attrsToUpdate Obj
 	return s.CreateObject(obj, NoConditions{})
 }
 
-func (s *storageFS) UpdateObject(bucketName, objectName string, metadata map[string]string) (StreamingObject, error) {
+func (s *storageFS) UpdateObject(bucketName, objectName string, attrsToUpdate ObjectAttrs) (StreamingObject, error) {
 	obj, err := s.GetObject(bucketName, objectName)
 	if err != nil {
 		return StreamingObject{}, err
 	}
 	defer obj.Close()
-	obj.Metadata = map[string]string{}
-	for k, v := range metadata {
-		obj.Metadata[k] = v
+
+	if attrsToUpdate.Metadata != nil {
+		obj.Metadata = map[string]string{}
 	}
+	obj.patch(attrsToUpdate)
 	obj.Generation = 0 // reset generation id
 	return s.CreateObject(obj, NoConditions{})
 }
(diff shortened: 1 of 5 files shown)


fsouza:
> I noticed that there are currently no tests for `UpdateObject`. I would be open to adding one or a few if they are needed.

I think it's because the Go SDK never invokes the update endpoint, so we'd need lower level tests built on top of the HTTP client. We have that for other endpoints though, so feel free to add it 😁

RachitSharma2001:
Sorry for the late push. I have added the test `TestServerClientObjectUpdateCustomTime` in `object_test.go`. Let me know if there is anything that should be added or changed.

internal/backend/fs.go
```diff
@@ -336,16 +336,14 @@ func (s *storageFS) PatchObject(bucketName, objectName string, attrsToUpdate Obj
...
 }
 
-func (s *storageFS) UpdateObject(bucketName, objectName string, metadata map[string]string) (StreamingObject, error) {
+func (s *storageFS) UpdateObject(bucketName, objectName string, attrsToUpdate ObjectAttrs) (StreamingObject, error) {
 	obj, err := s.GetObject(bucketName, objectName)
 	if err != nil {
 		return StreamingObject{}, err
 	}
 	defer obj.Close()
-	obj.Metadata = map[string]string{}
```
fsouza, line 345: Since this is supposed to reset Metadata, we should do something like:

```go
if attrsToUpdate.Metadata != nil {
    obj.Metadata = map[string]string{}
}
```

internal/backend/memory.go
```diff
@@ -310,16 +310,14 @@ func (s *storageMemory) PatchObject(bucketName, objectName string, attrsToUpdate
...
 
-// UpdateObject replaces an object metadata.
-func (s *storageMemory) UpdateObject(bucketName, objectName string, metadata map[string]string) (StreamingObject, error) {
+// UpdateObject replaces an object metadata, custom time, and acl.
+func (s *storageMemory) UpdateObject(bucketName, objectName string, attrsToUpdate ObjectAttrs) (StreamingObject, error) {
 	obj, err := s.GetObject(bucketName, objectName)
 	if err != nil {
 		return StreamingObject{}, err
 	}
-	obj.Metadata = map[string]string{}
```
fsouza, line 319: same as above

internal/backend/fs.go
RachitSharma2001: Ah yes that makes sense, since `obj.patch()` never resets metadata.

fsouza (approved):
Thank you very much for working on this!

Issue #1033, opened 2023-01-06
Unable to add files directly on the filesystem


drehelis:
Running `fake-gcs` with the following args:

```
docker run --rm --name fake-gcs-server -p 4443:4443 -v ${PWD}/examples1/data:/data fsouza/fake-gcs-server -scheme http -data /data -filesystem-root /data

curl "localhost:4443/storage/v1/b"
{"kind":"storage#buckets","items":[]}

mkdir examples1/data/{1..9}

curl "localhost:4443/storage/v1/b" | jq
  % Total    % Received % Xferd  Average Speed   Time    Time     Time  Current
                                 Dload  Upload   Total   Spent    Left  Speed
100   820  100   820    0     0  54503      0 --:--:-- --:--:-- --:--:-- 91111
{
  "kind": "storage#buckets",
  "items": [
    {
      "kind": "storage#bucket",
      "id": "1",
      "name": "1",
      "versioning": {},
      "location": "US-CENTRAL1"
    },
    {
      "kind": "storage#bucket",
      "id": "2",
      "name": "2",
      "versioning": {},
      "location": "US-CENTRAL1"
    },
   ...
   ...
   ... 
  ]
}

curl "localhost:4443/storage/v1/b/1/o"
{"kind":"storage#objects","items":[]}

touch examples1/data/1/dummy.txt
curl "localhost:4443/storage/v1/b/1/o"
{"error":{"code":404,"message":"Not Found","errors":null}}

restarting fake-gcs.

curl "localhost:4443/storage/v1/b/1/o"
{"kind":"storage#objects","items":[{"kind":"storage#object","name":"dummy.txt","id":"1/dummy.txt","bucket":"1","size":"0","contentType":"text/plain; charset=utf-8","crc32c":"AAAAAA==","md5Hash":"1B2M2Y8AsgTpgAmY7PhCfg==","etag":"\"1B2M2Y8AsgTpgAmY7PhCfg==\"","timeCreated":"2023-01-06T21:33:21.935431Z","updated":"2023-01-06T21:33:21.935455Z","generation":"1673040801935643"}]}


```

fsouza:
Yeah, that's not supported and can't really be supported. The format in the filesystem root is not regular folders and files, and it'll change further once that backend supports generations (#744).

One thing we could do is to add an internal endpoint for "reseeding" fake-gcs-server from a folder, like we do on process start.

drehelis:
[...]

PR #1079, opened 2023-02-19
Add Reseed Endpoint

RachitSharma2001:
Addresses #1033 

This adds a reseed endpoint at `_internal/reseed`

To reseed, the `/data` folder is reread fully and the server's backend in reinitialized. I fix the folder to be `/data` since this is also fixed in `internal/config.go` [here](https://github.com/fsouza/fake-gcs-server/blob/main/internal/config/config.go#L68). 

I also added a test on the helper function that the reseed endpoint uses.

Please let me know if there are any changes or things to add.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -20,11 +20,14 @@ import (
 	"net/http/httputil"
 	"net/textproto"
 	"net/url"
+	"os"
+	"path/filepath"
 	"strings"
 	"sync"
 
 	"cloud.google.com/go/storage"
 	"github.com/fsouza/fake-gcs-server/internal/backend"
+	"github.com/fsouza/fake-gcs-server/internal/checksum"
 	"github.com/fsouza/fake-gcs-server/internal/notification"
 	"github.com/gorilla/handlers"
 	"github.com/gorilla/mux"
@@ -260,6 +263,7 @@ func (s *Server) buildMuxer() {
 	// Internal / update server configuration
 	s.mux.Path("/_internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
 	s.mux.MatcherFunc(s.publicHostMatcher).Path("/_internal/config").Methods(http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.updateServerConfig))
+	s.mux.Path("/_internal/reseed").Methods(http.MethodPut, http.MethodPost).HandlerFunc(jsonToHTTPHandler(s.reseedServer))
 	// Internal - end
 
 	bucketHost := fmt.Sprintf("{bucketName}.%s", s.publicHost)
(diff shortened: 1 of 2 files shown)


fakestorage/server.go
```diff
@@ -287,6 +291,85 @@ func (s *Server) buildMuxer() {
 	s.mux.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods(http.MethodPost, http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.insertObject))
 }
 
+func (s *Server) reseedServer(r *http.Request) jsonResponse {
+	folder := "/data"
+	initialObjects, emptyBuckets := generateObjectsFromFiles(folder)
+	for _, bucketName := range emptyBuckets {
+		s.CreateBucketWithOpts(CreateBucketOpts{Name: bucketName})
+	}
```
fsouza, line 299: hmm this is tricky because if we reassign the backend below, those buckets are going to be gone. Perhaps the backend constructor could take a list of empty buckets too?

RachitSharma2001: That is a good point. However, when I run the following script, where three empty buckets are created before the reseed, it seems like those buckets are detected:
```
docker run -d --name fake-gcs-server -p 4443:4443 -v ${PWD}/all_files:/data fake-server-reseed
mkdir ./all_files/bucket2
mkdir ./all_files/bucket3
mkdir ./all_files/bucket4

echo "Before Reseed:"
for i in {2..4}; do
    echo "Contents of bucket$i"
    curl -k "https://localhost:4443/storage/v1/b/bucket$i/o"
done

echo "Hitting the reseed endpoint:"
curl -k -X POST "https://localhost:4443/_internal/reseed"
echo ""

echo "After Reseed:"
for i in {2..4}
do
    echo "Contents of bucket$i"
    curl -k "https://localhost:4443/storage/v1/b/bucket$i/o"
done
```

Output:
```
...
After Reseed:
Contents of bucket2
{"kind":"storage#objects","items":[]}
Contents of bucket3
{"kind":"storage#objects","items":[]}
Contents of bucket4
{"kind":"storage#objects","items":[]}
```

However, I definitely see the merit of passing in the `emptyBuckets` into the backend `NewStorageFS` and `NewStorageMemory` functions as that would also simplify `main.go`. I can push a change for this.

RachitSharma2001:
I have updated the code so that the backend server does not replace the empty buckets. To do this, I create the empty buckets after reassigning the backend server (see lines 310-313 [here](https://github.com/fsouza/fake-gcs-server/pull/1079/files#diff-913ea60f158d61cac04a406321b4ef4a7d27d665a6b38da65a3d2fe6f05d832c)). Let me know if there are any other things that should change in this implementation.

fsouza (approved):
Thank you very much for working on this and apologies for taking so long to get to it.

PR #1112, opened 2023-03-30
Fix/fix some directory issues

mahmoudbahaa:
-  add updated field to bucket response and Model
- better support directories in filesystem //TODO: do the same for memory
- use removeAll instead of remove when deleting a file/directory
- introduce rename function that is used in rewrite to rename directories properly

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -713,25 +713,20 @@ func (s *Server) rewriteObject(r *http.Request) jsonResponse {
 	}
 
 	dstBucket := vars["destinationBucket"]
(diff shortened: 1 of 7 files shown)


mahmoudbahaa:
@fsouza any updates or changes needed for the PR ?

mahmoudbahaa:
@fsouza updated PR is here https://github.com/fsouza/fake-gcs-server/pull/1133

fsouza:
@mahmoudbahaa thank you! I'll push back on the deadline a bit, but I plan to take a look at that PR before Wednesday!

PR #1115, opened 2023-04-02
feat(xml): add support for list objects

goober:
This pull request aims to add support for listing objects according to the [Google Cloud Storage XML API](https://cloud.google.com/storage/docs/xml-api/get-bucket-list)


fsouza (commented):
Hey, thanks for contributing. I'm planning to have a look at this later today, can you rebase to fix the conflict? Thank you!

fakestorage/object.go
```diff
@@ -323,38 +325,63 @@ type ListOptions struct {
 	StartOffset              string
 	EndOffset                string
 	IncludeTrailingDelimiter bool
+
+	PageToken string
+	MaxKeys   *int
```
fsouza, line 330: Do you think you could send the pagination changes separately? Also, ideally we'd introduce them without breaking the signature of `ListObjects` and `ListObjectsWithOptions` (which maybe means we need a third method, which is definitely not great). We can make that method private though and have `ListObjects` and `ListObjectsWithOptions` always return all objects.

fsouza (commented):
I pushed some changes to your branch to get the PR green, but feel free to override them.

I think this looks good overall, but I'd like to have the pagination change as a separate PR or available only through the HTTP API, we don't need to expose pagination as part of the library surface (like let's implement it in an unexported method that gets used by `ListObjectsWithOptions`).

goober:
Thanks for the review @fsouza, and I agree that we should strive for not breaking the current api. So I have removed everything that is related to pagination in the current PR. And then we could have another look at that in a separate PR later on, if necessary.

fsouza (approved):
Thank you!

PR #1126, opened 2023-04-08
fix: make api copyTo return an object resource

### LINKED ISSUE #1116: API copy Object should return an object resource on success.
Hi!
First of all, thank you for developing this. It's an invaluable tool for local development 👏

## Problem

I found that the copyTo and rewriteTo API is using the same rewriteObject function.
The rewriteTo API is OK but copyTo API response the same make client library didn't work.
Can you update the rewriteObject function to return an object resource if request to copyTo API.

https://github.com/fsouza/fake-gcs-server/blob/e610de42cc1616d910ce74f21cf12b69e827aa5c/fakestorage/server.go#L257-L258

<details><summary>You can check the docs of GCS for more information:</summary>
<p>

- https://cloud.google.com/storage/docs/json_api/v1/objects/rewrite#response
- https://cloud.google.com/storage/docs/json_api/v1/objects/copy#response

</p>
</details> 

## Proposed solution

Check the path if it is copyTo request return an object resource.

Thank you!

CanLe295:
fix #1116 

- api copy object should return an  object resource

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -731,6 +731,9 @@ func (s *Server) rewriteObject(r *http.Request) jsonResponse {
 	}
 	defer created.Close()
 
+	if vars["copyType"] == "copyTo" {
+		return jsonResponse{data: newObjectResponse(created.ObjectAttrs)}
+	}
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thank you!

PR #1133, opened 2023-04-15
Fix/fix some directory issues

mahmoudbahaa:
add updated field to bucket response and Model
better support directories in filesystem //TODO: do the same for memory

### DIFF
--- a/fakestorage/response.go
+++ b/fakestorage/response.go
@@ -42,6 +42,7 @@ type bucketResponse struct {
 	Name        string            `json:"name"`
 	Versioning  *bucketVersioning `json:"versioning,omitempty"`
 	TimeCreated string            `json:"timeCreated,omitempty"`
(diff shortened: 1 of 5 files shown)


fsouza (commented):
Hmm, I don't think this will work. When you create an object called a/b/c/d/e.txt, that should be the only object. With this change, we're considering all the sub-directories individual objects (you can take a look at the test failures).

Unfortunately, the current filesystem backend can't really support objects that are just a directory 😞 We really need a new filesystem-based backend that can store fake-gcs-server specific metadata about objects, this would enable us to support objects that look like directories and also support generations in the fs backend.

Putting generations aside, we could potentially leverage the metadataHandler to detect whether a directory is an object or not (we'd need to extend the metadataHandler interface I think, and maybe add some special attributes to that).

PR #1164, opened 2023-05-12
Support XML API (#331)

### LINKED ISSUE #331: Consider XML support for object PUT/GET
First, thank you for writing this!

Some client libraries (e.g. [C++](https://github.com/googleapis/google-cloud-cpp)) automatically switch to XML for some uploads and downloads.  You may want to consider supporting the XML APIs to insert and download objects.  In case it helps, our testbench implements these. These are the entry points, hopefully the code is not so terrible that you can follow it from there:

https://github.com/googleapis/google-cloud-cpp/blob/abc8e9c3e95c561924a8ae567012440152d097b6/google/cloud/storage/testbench/testbench.py#L50-L67

tustvold:
Closes #331

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -276,6 +276,9 @@ func (s *Server) buildMuxer() {
 	for _, r := range xmlApiRouters {
 		r.Path("/").Methods(http.MethodGet).HandlerFunc(xmlToHTTPHandler(s.xmlListObjects))
 		r.Path("").Methods(http.MethodGet).HandlerFunc(xmlToHTTPHandler(s.xmlListObjects))
+		r.Path("/{objectName:.+}").Methods(http.MethodPut).HandlerFunc(xmlToHTTPHandler(s.xmlPutObject))
+		r.Path("/{objectName:.+}").Methods(http.MethodGet, http.MethodHead).HandlerFunc(s.downloadObject)
+		r.Path("/{objectName:.+}").Methods(http.MethodDelete).HandlerFunc(xmlToHTTPHandler(s.xmlDeleteObject))
 	}
 
 	bucketHost := fmt.Sprintf("{bucketName}.%s", s.publicHost)
@@ -295,12 +298,6 @@ func (s *Server) buildMuxer() {
 	// Form Uploads
 	handler.Host(s.publicHost).Path("/{bucketName}").MatcherFunc(matchFormData).Methods(http.MethodPost, http.MethodPut).HandlerFunc(xmlToHTTPHandler(s.insertFormObject))
 	handler.Host(bucketHost).MatcherFunc(matchFormData).Methods(http.MethodPost, http.MethodPut).HandlerFunc(xmlToHTTPHandler(s.insertFormObject))
-
-	// Signed URLs (upload and download)
-	handler.MatcherFunc(s.publicHostMatcher).Path("/{bucketName}/{objectName:.+}").Methods(http.MethodPost, http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.insertObject))
-	handler.MatcherFunc(s.publicHostMatcher).Path("/{bucketName}/{objectName:.+}").Methods(http.MethodGet, http.MethodHead).HandlerFunc(s.getObject)
-	handler.Host(bucketHost).Path("/{objectName:.+}").Methods(http.MethodPost, http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.insertObject))
-	handler.Host("{bucketName:.+}").Path("/{objectName:.+}").Methods(http.MethodPost, http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.insertObject))
 
 	s.handler = handler
 }
--- a/fakestorage/upload_test.go
+++ b/fakestorage/upload_test.go
@@ -10,7 +10,6 @@ import (
 	"context"
 	"crypto/tls"
 	"encoding/binary"
-	"encoding/json"
 	"fmt"
 	"io"
 	"mime/multipart"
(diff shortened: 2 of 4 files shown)


fakestorage/upload_test.go
```diff
@@ -596,23 +592,6 @@ func TestServerClientSignedUploadBucketCNAME(t *testing.T) {
 	if resp.StatusCode != http.StatusOK {
 		t.Errorf("wrong status returned\nwant %d\ngot  %d", http.StatusOK, resp.StatusCode)
 	}
-	data, err := io.ReadAll(resp.Body)
```
tustvold, line 599: This was incorrect, the XML APIs should not return a body and certainly shouldn't return JSON

fakestorage/server.go
```diff
@@ -296,12 +299,6 @@ func (s *Server) buildMuxer() {
 	handler.Host(s.publicHost).Path("/{bucketName}").MatcherFunc(matchFormData).Methods(http.MethodPost, http.MethodPut).HandlerFunc(xmlToHTTPHandler(s.insertFormObject))
 	handler.Host(bucketHost).MatcherFunc(matchFormData).Methods(http.MethodPost, http.MethodPut).HandlerFunc(xmlToHTTPHandler(s.insertFormObject))
 
-	// Signed URLs (upload and download)
-	handler.MatcherFunc(s.publicHostMatcher).Path("/{bucketName}/{objectName:.+}").Methods(http.MethodPost, http.MethodPut).HandlerFunc(jsonToHTTPHandler(s.insertObject))
```
tustvold, line 300: This can't have ever worked as insertObject expects query parameters encoding the object name, etc...

fakestorage/upload_test.go
fsouza: 🤔 Oh are signed uploads powered by the XML API? Interesting!

fakestorage/server.go
fsouza: `insertObject` does not necessarily read the object name from query parameters: https://github.com/fsouza/fake-gcs-server/blob/d7832ecd0fab51f1e9047f53375b694aec31e29f/fakestorage/upload.go#L85-L95

I need to take a deeper look at this, but curious if you're basing on comment on the docs? There are many undocumented things in the GCS API that we run into when someone tries to integrate with some different SDK, and I've ironically done a poor job in documenting those in fake-gcs-server 🙈

tustvold: Indeed, insertObject has a peculiar special case that uses `X-Goog-Algorithm` to detect a signed URL and then falls back to implementing something that looks like the XML API. With this PR that can probably be removed :thinking:

fsouza: Yeah, absolutely. insertObject definitely looks like a little monster doing too many things. Thank you very much! I'll go over the PR in details later today.

ramir-savvy:
Hi, bumping this one - any plans to fix and merge it soon? Thanks

ramir-savvy:
Thank you @fsouza. I have opened #1215 which better suits my needs.

tustvold:
Just wondering what the status of this is, the lack of support for the XML APIs is causing some upstream confusion - https://github.com/apache/arrow-rs/issues/5263

Slach:
=( ohh, if could be a great functionality which allow us test GCS over S3 in https://github.com/Altinity/clickhouse-backup/issues/695

nickpresta:
@tustvold Is this something still on your radar?

tustvold:
We've been using this branch to run integration tests for the last year. I'd be happy to see this incorporated but that's out of my hands 😅

fsouza:
Yeah I think the code overall looks good, but I'd need conflicts resolved and CI to be green before merging. I can give it a shot myself, but I don't expect to have any time to work on this in the next 2-3 weeks.

Slach:
@fsouza if i merge conclicts and create new PR, will it merge?

adriangb:
it would be great to get this functionality included!

adriangb:
I tried using this with a mix of Rust's `object_store` (via `deltalake`) and `google-cloud-storage` in Python and can't get them to both work:

```python
import os

os.environ['STORAGE_EMULATOR_HOST'] = 'http://localhost:4443'
os.environ['GOOGLE_SERVICE_ACCOUNT_KEY'] = '{"gcs_base_url": "http://localhost:4443", "disable_oauth": true, "client_email": "", "private_key": "", "private_key_id": ""}'

import pyarrow
from deltalake import DeltaTable
from google.cloud import storage

client = storage.Client()

bucket = client.bucket('test-bucket')  # type: ignore

print(list(bucket.list_blobs()))

bucket.blob('test.txt').upload_from_string(b'Hello, world!')

print(list(bucket.list_blobs()))

blob = bucket.blob('test.txt')

print(blob.download_as_string())

DeltaTable.create('gs://test-bucket', schema=pyarrow.schema([('id', pyarrow.int64())]), mode='overwrite')
```

Running `docker run -it -p 4443:4443 -v ./gcs_data:/data tustvold/fake-gcs-server -scheme http -public-host localhost:4443 -log-level debug -filesystem-root=/data` makes the deltalake / object_store part work but the google sdk part fails at `blob.download_as_string()` with `'Request failed with status code', 404, 'Expected one of', <HTTPStatus.OK: 200>, <HTTPStatus.PARTIAL_CONTENT: 206>` and I see `"GET /download/storage/v1/b/test-bucket/o/test.txt?alt=media HTTP/1.1" 404 10` in the logs.

Switching to running `docker run -it -p 4443:4443 -v ./gcs_data:/data tustvold/fake-gcs-server -scheme http -public-host http://localhost:4443 -log-level debug -filesystem-root=/data` (so adding `http://` to `-public-host`) makes the google sdk part work but the object_store part fail with `OSError: Generic GCS error: Error performing list request: Client error with status 404 Not Found: Not Found` and I see `"GET /test%2Dbucket?list-type=2&prefix=_delta_log%2F HTTP/1.1" 404 10` in the logs.

hunterp:
I just ran into this issue today.  @fsouza it would be really helpful if this could still get merged!

dreamtalen:
+1, I encountered the same issue while testing with the `object_store` crate. Looking forward to seeing it merged.

LarsAlmgren:
This PR seems stale @fsouza so I've fixed the conflicts and pushed to another branch PTAL - https://github.com/fsouza/fake-gcs-server/pull/2092

PR #1167, opened 2023-05-16
Add bucket filter for notification events

mccalltd:
Thanks for this project! 

I found the current object level filter for bucket notifications insufficient for my usage.

This PR extends the notification event config to accept a bucket filter. I could change this to accept a CSV of bucket names instead, and would be happy to do so if you thought that would be better. Left it simple to start.


fsouza (approved):
Thanks for contributing!

Issue #217, opened 2020-04-19
gsutil cp bad request


jonasfugedi:
I need to test some bash scripts which run gsutil but I keep getting errors when trying to create objects using gsutil. The scenario I run is basically:

docker  run --name fake-gcs-server -p 4443:4443 fsouza/fake-gcs-server

gsutil -o "Credentials:gs_json_host=0.0.0.0" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" mb "gs://test"

gsutil -o "Credentials:gs_json_host=0.0.0.0" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" ls "gs://test"

echo "Hello" | gsutil  -o "Credentials:gs_json_host=0.0.0.0" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" cp - "gs://test/hello.txt"

Copying from <STDIN>...
ResumableUploadStartOverException: 404 Bad Request

gsutil  -o "Credentials:gs_json_host=0.0.0.0" -o "Credentials:gs_json_port=4443" -o "Boto:https_validate_certificates=False" cp ./tmp/funny-memes-81.jpg  "gs://test/"

Copying file://./tmp/funny-memes-81.jpg [Content-Type=image/jpeg]...
BadRequestException: 400 Bad Request

fsouza:
Hey @jonasfugedi, thanks for opening this issue. What do you see in the server logs?

jonasfugedi:
The logs did not give me any good clues. Is there any debug flag I can enable to get more details?

`docker run --name fake-gcs-server -p 4443:4443 fsouza/fake-gcs-server
Unable to find image 'fsouza/fake-gcs-server:latest' locally
latest: Pulling from fsouza/fake-gcs-server
aad63a933944: Pull complete
cff4c0a9b296: Pull complete
Digest: sha256:c4388cf0e1d5570ba2600575aebb909bec827c975b632533f6af1d8c448f9c40
Status: Downloaded newer image for fsouza/fake-gcs-server:latest
time="2020-04-19T16:45:42Z" level=info msg="couldn't load any objects or buckets from \"/data\", starting empty"
time="2020-04-19T16:45:42Z" level=info msg="server started at https://[::]:4443"
[...]

PR #1182, opened 2023-05-26
multipart uploads: Hack to support gsutil cp

ekimekim:
`gsutil` sends an invalid multipart boundary param, which golang's `mime.ParseMediaType` correctly rejects.
However, the real GCS evidently does not reject this, so in order to make `gsutil` work we need to support it.

In particular, `gsutil` sends a boundary param that is quoted using single-quotes when it should be using double-quotes.
In cases where the param is definitely invalid (so we're guarenteed not to break any valid values), we replace all single-quotes with double-quotes to produce the intended meaning.

Upstream bug: https://github.com/GoogleCloudPlatform/gsutil/issues/1466

Fixes https://github.com/fsouza/fake-gcs-server/issues/217
The above issue found this bug and worked around it, but found another problem and never submitted the fix.
It seems that in the intervening time, the other problem was fixed independently, and I can successfully `gsutil cp` using this code.

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -315,7 +315,7 @@ func (s *Server) multipartUpload(bucketName string, r *http.Request) jsonRespons
 	// If we see a string like "boundary='=", which is always invalid anyway,
 	// attempt to rescue the situation by converting all ' to ".
 	if strings.Contains(requestContentType, "boundary='=") {
-		requestContentType = strings.ReplaceAll(requestContentType, "'", "\"")
+		requestContentType = strings.ReplaceAll(requestContentType, "'", `"`)
 	}
 	_, params, err := mime.ParseMediaType(requestContentType)
 	if err != nil {
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -306,7 +306,18 @@ func getObjectACL(predefinedACL string) []storage.ACLRule {
 
 func (s *Server) multipartUpload(bucketName string, r *http.Request) jsonResponse {
 	defer r.Body.Close()
-	_, params, err := mime.ParseMediaType(r.Header.Get(contentTypeHeader))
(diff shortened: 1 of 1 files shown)


ekimekim:
hm, now I think about it more, it might be safer to only do the `'` -> `"` transform if we first try and fail to parse the mime type. right now, a boundary value like `boundary="boundary='="` would do the wrong thing.

fakestorage/upload.go
```diff
@@ -306,7 +306,18 @@ func getObjectACL(predefinedACL string) []storage.ACLRule {
...
+	requestContentType := r.Header.Get(contentTypeHeader)
+	// gsutil is observed to submit incorrectly-quoted bounary strings
+	// like: boundary='===============5900997287163282353=='
+	// See https://github.com/GoogleCloudPlatform/gsutil/issues/1466
+	// Having an "=" character in the boundary param requires the value be quoted,
+	// but ' is not a quote char, " is.
+	// If we see a string like "boundary='=", which is always invalid anyway,
+	// attempt to rescue the situation by converting all ' to ".
+	if strings.Contains(requestContentType, "boundary='=") {
+		requestContentType = strings.ReplaceAll(requestContentType, "'", "\"")
```
fsouza, line 318: nit: can use ` to avoid having to escape

```suggestion
		requestContentType = strings.ReplaceAll(requestContentType, "'", `"`)
```

fsouza (commented):
Thanks for fixing this!

fsouza (approved):

fsouza:
> hm, now I think about it more, it might be safer to only do the ' -> " transform if we first try and fail to parse the mime type. right now, a boundary value like boundary="boundary='=" would do the wrong thing.

Yeah, I wonder if we need something more specific, like a regexp replace that is specific to `boundary='(.+)'` or something like that.

PR #1215, opened 2023-06-15
Add scheme option to bind to both HTTP and HTTPS

ramir-savvy:
We have a use case where we need the same server to run in both HTTP and HTTPS.
This PR starts two servers when needed (the default single HTTPS server is kept).
It is controlled by adding/changing the following command line arguments:
1. Existing `scheme` argument can now get the value `both`
2. New `port-http` argument, defaults to `8080`, is used to bind HTTP port only when `scheme` is set to `both`. To clarify: to maintain backwards compatibility, when the `scheme` is set to `http`, this argument isn't used. The previous `port` will be used to bind HTTPS.

Added/updated tests
Updated README


fsouza:
> We have a use case where we need the same server to run in both HTTP and HTTPS.

Interesting. Can you share more details on that use case? Like, why you need the server to bind on both?

ramir-savvy:
@fsouza, we have an internal testing setup that is running with http because it was easier to setup. We want to use gsutil in that setup, and it supports https only.

main.go
```diff
@@ -81,7 +70,27 @@ func main() {
...
+
+	logger := logrus.New()
+	logger.SetLevel(cfg.LogLevel)
+
+	if cfg.Scheme != "both" {
+		startServer(logger, &cfg, cfg.Scheme)
+	} else {
+		startServer(logger, &cfg, "http")
+		startServer(logger, &cfg, "https")
+	}
```
fsouza, line 93: Can we setup two listeners that point to the same server instance? I feel like that would make the most sense.

fsouza:
@ramir-savvy gotcha, thanks for clarifying!

main.go
ramir-savvy: @fsouza , I can do that, although I am not sure there is a big penalty when having two servers.
Started implementing but ran into lister code inside `NewServerWithOptions` which complicates a bit the transition to 2 listeners with a single server:
https://github.com/ramir-savvy/fake-gcs-server/blob/ea0ba662c5381c1317c13151f231e361cb7e51e9/fakestorage/server.go#L163

`options.NoListener` is set to `true` by default and only set to `false` in the tests.
Can you clarify its use?
If the `NewServerWithOptions` function just returned in that line instead of starting listeners for tests it would be easier.
Or maybe keep it as it is and ignore that in the server run because it is relevant just for tests?

fsouza: > although I am not sure there is a big penalty when having two servers

I was under the assumption that you wanted both servers to serve the same underlying buckets and objects? Without that, the memory backend will not have the same underlying objects, and I'd prefer that we don't diverge the capabilities of the backends here.

We can also decouple the backend from the server (i.e. support taking a backend instance in the server constructor), and in that case having two servers won't matter (similar to how the grpc server takes the backend instance).

ramir-savvy: @fsouza , you are right of course, I didn't take that into account.

I have pushed a commit that changes the code to have 2 listeners and a single server.
I looked at decoupling the backend from the server but it didn't seem to be so trivial.
Let me know what you think. Thanks

fsouza (commented):
Can you take a look at the test failure? Thanks!

ramir-savvy:
Apologies @fsouza !
First: I was focused on "both" and neglected testing without the `scheme` param - fixed now.
Second: I didn't notice that you ran the tests, so I wasn't aware of the failure till you pinged me - Can you ping me when you start rerunning the tests?
Thanks

internal/config/config.go
```diff
@@ -173,14 +178,18 @@ func (c *Config) ToFakeGcsOptions() fakestorage.Options {
 			}
 		}
 	}
+	port := c.Port
+	if scheme == "http" {
+		port = c.PortHTTP
+	}
```
fsouza, line 184: Won't this be a breaking change? Like, if I'm currently using fake-gcs-server like this:

```
fake-gcs-server -scheme http -port 8080
```

The server will now start on port `8000` instead of `8080`.

I know this is fairly complicated, but could we make it so:

1. `port` and `port-http` no longer have a default at the flag definition. We'll handle it ourselves
2. if the user specifies `-scheme http` and `-port`, we use that port (keep current behavior)
3. if the user specifies just `-scheme http`, we use the default `-port` (keep current behavior)
4. if the user specifies just `-scheme both`, we'll use the default `-port` and `-port-http`
5. if the user doesn't provide any flags, we'll use the default `-port` and `-scheme https`

Perhaps we can turn each of the 5 examples in test cases for the function/core logic that handles this.

We may be able to implement this custom logic with a custom implementation of [flag.Value](https://pkg.go.dev/flag#Value), but it sounds tricky as it'd need to be aware of the value of another flag. The biggest advantage of using a custom flag.Value is differentiating between the user not providing the flag and setting it to `0` (I'm not super concerned about that though - right now if the user sets it to `0`, we start the server in a random port, but never print that port to the user, so who cares? heh)

fsouza (commented):
Thank you! It looks like some examples are failing due to the comment I left about existing usages of `-scheme http -port <something>`.

Let me know what you think!

internal/config/config.go
ramir-savvy: You are correct. I overlooked the http port in my last fix and broke the behavior, that I previously maintained. Thanks for catching that!

I implemented the `port` and `port-http` without defaults in the `flag` API but set it myself in the code if their value is 0. As you suggest, I also think that no one uses it today with port 0, so I don't think it is a breaking change, so I didn't implement the custom `flag.Value`.
I have an idea of how to implement it, maybe using `flag.Visit`, and I will do it if you think it is needed.
Please let me know. 

Will work on adding the test cases soon.

ramir-savvy: @fsouza 
Added unit tests to cover the examples you listed and few more to be on the safe side. The tests found an issue with `both` scheme in the `cfg.externalURL` config - fixed.
Added 2 more runs to the github action to cover `both` scheme with default and non-default ports.

Let me know what you think. Thanks

fsouza (approved):
Thank you!

ramir-savvy:
@fsouza, I just learned that after all there is a breaking change with this PR.
Before this PR, when the server was started with `-scheme http` **without** specifying `-port`, the default port was `443` (even though it was http). This PR changed it to `8000`.
Sorry about that.
I will submit a PR soon to revert this behavior.

PR #1227, opened 2023-06-26
Send X-Goog-Stored-Content-Encoding header

### LINKED ISSUE #1226: bug: invalid CRC computed
In previous versions of the GCS storage client (`cloud.google.com/go/storage`, up to v1.29.0), there was a parsing bug that caused the client to skip CRC validation.

However, in v1.30.0, [a fix](https://github.com/googleapis/google-cloud-go/commit/574a86c614445f8c3f5a54446820df774c31cd47) for that bug was released, and since then I've received this error while reading:

```
storage: bad CRC on read: got 1077016486, want 2266108419
```

My hunch is the CRC calculation in `fsouza/fake-gcs-server` is incorrect (compared to how GCS does it); when the feature was added (https://github.com/fsouza/fake-gcs-server/pull/307), this comment in the description stands out:

> I compared the results against actually uploading to GCS (and using `gsutil hash -m`) and they differ for some reason. Whilst this is somewhat annoying it isn't a problem for me as all I need is something consistent.

I think this wasn't a problem then, since the header wasn't used, but now that `cloud.google.com/go/storage` has fixed that bug this is now a problem!

/cc @antonydenyer @fsouza

hairyhenderson:
Fixes #1226

I've also fixed some unit tests that were passing but shouldn't have been.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -887,6 +887,13 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 		if obj.ContentEncoding != "" && !transcoded {
 			w.Header().Set("Content-Encoding", obj.ContentEncoding)
 		}
+		// X-Goog-Stored-Content-Encoding must be set to the original encoding,
+		// defaulting to "identity" if no encoding was set.
+		storedContentEncoding := "identity"
+		if obj.ContentEncoding != "" {
+			storedContentEncoding = obj.ContentEncoding
+		}
+		w.Header().Set("X-Goog-Stored-Content-Encoding", storedContentEncoding)
 	}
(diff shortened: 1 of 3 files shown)


hairyhenderson:
@fsouza can I get this reviewed? I've been running this patch in production for a few hours and it seems like it's working fine.

fsouza (approved):
Thank you!

gnagel:
@fsouza - Will there be a release with this update soon?

TY!

fsouza:
@gnagel thanks for checking in. I'm going through some of the open PRs and will push a new release later today!

gnagel:
Awesome, ty so much 😁

PR #1254, opened 2023-07-17
Disallow composing more than 32 objects at once

manuteleco:
The Google Cloud Storage API does not allow it either (see [documentation][1]). Trying to compose more than 32 objects at once qualifies as a 400 BadRequest. This is the body that GCS returns in such a case:

```json
{
  "error": {
    "code": 400,
    "message": "The number of source components provided (33) exceeds the maximum (32)",
    "errors": [
      {
        "message": "The number of source components provided (33) exceeds the maximum (32)",
        "domain": "global",
        "reason": "invalid"
      }
    ]
  }
}
```

Note that GCS supports composing more than 32 objects into one. But it requires multiple `compose` calls, each one composing at most 32 objects in one go.

[1]: https://cloud.google.com/storage/docs/json_api/v1/objects/compose

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -1121,6 +1121,14 @@ func (s *Server) composeObject(r *http.Request) jsonResponse {
 		}
 	}
 
+	const maxComposeObjects = 32
+	if len(composeRequest.SourceObjects) > maxComposeObjects {
+		return jsonResponse{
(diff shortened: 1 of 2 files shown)


fsouza (commented):
Thanks for contributing! Can you add a test?

fsouza (approved):
Thank you!

PR #1272, opened 2023-08-01
Restore 4443 as default port for '-scheme http' as before PR #1215

ramir-savvy:
Fix breaking change caused by PR #1215:
Restore the default port that was used for `-scheme http` before the integration that PR


ramir-savvy:
@fsouza, here is the fix. Thanks!

fsouza (approved):

fsouza:
Thanks for the quick fix, I push 1.47.1 shortly.

PR #1361, opened 2023-10-12
Updated python example

j-antunes:
I tried the python example and it is not working out of the box. The main issue that I was facing is:

```
google.api_core.exceptions.BadRequest: 400 GET https://storage.googleapis.com/storage/v1/b?project=test&projection=noAcl&prettyPrint=false: Invalid project number: 0. Project id: 0 is invalid or not found
```

I think the issue has to do with not setting the `api_endpoint` for the docker container.

The solution is to add
```
    client_options={"api_endpoint": "http://localhost:4443"},
```

Apart from that I added some instructions as to how to run this example.

### DIFF
--- a/examples/python/python.py
+++ b/examples/python/python.py
@@ -2,6 +2,16 @@
 # Use of this source code is governed by a BSD-style
 # license that can be found in the LICENSE file.
 
+# How to run this example
+# 1 - Build the docker image by running the command "docker build -t fsouza/fake-gcs-server ."
+# 2 - Start the docker container: "docker run -d --name fake-gcs-server -p 4443:4443 -v ${PWD}/examples/data:/data fsouza/fake-gcs-server -scheme http"
+# 3 - Check if it's working by running: "curl http://0.0.0.0:4443/storage/v1/b"
+# 4 - Create a python virtual enviroment (Ex: python -m .venv venv)
+# 5 - Source the env (source .venv/bin/activate)
+# 7 - Go to the following directory examples/python: (cd examples/python)
+# 6 - Install requirements: "pip install -r requirements.txt" and "pip install -r requirements.in"
+# 7 - Run this script
+
 import tempfile
 
 from google.auth.credentials import AnonymousCredentials
@@ -10,6 +20,9 @@ from google.cloud import storage
 client = storage.Client(
     credentials=AnonymousCredentials(),
     project="test",
+    # This endpoint assumes that you are using the default port 4443 from the container.
+    # If you are using a different port you need to update this endpoint
+    client_options={"api_endpoint": "http://localhost:4443"},
 )
 
 # List the Buckets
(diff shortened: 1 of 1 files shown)


fsouza (commented):
The instructions on how to run the example can be derived from the CI script that executes the example in GHA. Can you update the comment to reflect that and undo the `client_options` step? Otherwise, this breaks CI.

j-antunes:
@fsouza - I followed you suggestion and update my example.

I added this line
`os.environ["STORAGE_EMULATOR_HOST"] = os.environ.get(
    "STORAGE_EMULATOR_HOST", "http://localhost:4443"
)`

That should help to make this example more clear without breaking your CI. This tries to get the value of the STORAGE_EMULATOR_HOST environment variable. If it doesn't exist, it provides a default value of "http://localhost:4443/".

When you run your CI, you set this value, so it will use localhost:8080, if running just the python script with the instructions it will set to localhost:4443

examples/python/python.py
```diff
@@ -2,14 +2,38 @@
...
+import os
 
 from google.auth.credentials import AnonymousCredentials
 from google.cloud import storage
 
+# This endpoint assumes that you are using the default port 4443 from the container.
+# If you are using a different port you need to update this endpoint
+os.environ["STORAGE_EMULATOR_HOST"] = os.environ.get(
+    "STORAGE_EMULATOR_HOST", "http://localhost:4443"
+)
```
fsouza, line 28:
```suggestion
# This endpoint assumes that you are using the default port 4443 from the container.
# If you are using a different port, please set the environment variable STORAGE_EMULATOR_HOST.
os.environ.setdefault("STORAGE_EMULATOR_HOST", "http://localhost:4443")
```

fsouza (approved):
Thanks!

examples/python/python.py
```diff
@@ -2,14 +2,36 @@
...
 # license that can be found in the LICENSE file.
 
+# How to run this example
+# 1 - Build the docker image by running the command "docker build -t fsouza/fake-gcs-server ."
+# 2 - Start the docker container: "docker run -d --name fake-gcs-server -p 4443:4443 -v ${PWD}/examples/data:/data fsouza/fake-gcs-server -scheme http"
+# 3 - Check if it's working by running: "curl http://0.0.0.0:4443/storage/v1/b"
+# 4 - Create a python virtual enviroment (Ex: python -m .venv venv)
+# 5 - Source the env (source .venv/bin/activate)
+# 7 - Go to the following directory examples/python: (cd examples/python)
+# 6 - Install requirements: "pip install -r requirements.txt" and "pip install -r requirements.in"
```
fsouza, line 12: requirements.in should not be used

```suggestion
# 6 - Install requirements: "pip install -r requirements.txt"
```

fsouza (approved):
Thanks!

Issue #273, opened 2020-06-18
Patch call with new content-type does nothing


ex-nerd:
As far as I can see in the fake-gcs-server code, contentType is only set when the object is created, and can't be updated.

I'm using the official python gcs library, basically as follows:

```python
# init
from google.cloud import storage
client = BigMessyConnectionToFakeGcsServer(storage.Client)
bucket = client.bucket("my-test-bucket")
key = "test-key"

# upload some data
blob = bucket.blob(key)
blob.upload_from_string(b"some data", content_type="application/octet-stream")

# now fix the content type
blob = bucket.get_blob(key)
blob.content_type = "text/plain"
blob.patch()
```

The content-type value returned from the API request (buried in google's `patch()` method definition) is the original unchanged `application/octet-stream`, as is the content-type when trying to download the blob.

Debugging into google's code, this is sending `{"contentType": "application/some-new-type"}` along with the API request as the body content, which seems to be completely ignored by `Server.patchObject` unmarshaling into its `metadata` struct.

corytheboyd-smartsheet:
FWIW I think I just ran into this same issue. Wrote the integration test for our feature that updates an object's content-type, and was confused about what I was doing wrong with what I though was a very simple implementation 😅

PR #1430, opened 2023-12-15
Fixed patchObject ignoring ContentType and ContentEncoding

UnAfraid:
I was trying to update object's ContentType and noticed that only metadata was updated, but not the ContentType upon looking at the code i noticed that its ignored entirely.

Most likely fixes https://github.com/fsouza/fake-gcs-server/issues/273

### DIFF
--- a/internal/grpc/server.go
+++ b/internal/grpc/server.go
@@ -44,7 +44,6 @@ func (g *Server) GetBucket(ctx context.Context, req *pb.GetBucketRequest) (*pb.B
 		TimeCreated:           timestamppb.New(bucket.TimeCreated),
 	}
 	return grpc_bucket, nil
-	///return GetBucketFromBackend(g.backend, req.Bucket)
 }
 
 func (g *Server) DeleteBucket(ctx context.Context, req *pb.DeleteBucketRequest) (*pb.Empty, error) {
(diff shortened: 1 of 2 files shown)


internal/grpc/server.go
```diff
@@ -44,7 +44,7 @@ func (g *Server) GetBucket(ctx context.Context, req *pb.GetBucketRequest) (*pb.B
 		TimeCreated:           timestamppb.New(bucket.TimeCreated),
 	}
 	return grpc_bucket, nil
-	///return GetBucketFromBackend(g.backend, req.Bucket)
+	// /return GetBucketFromBackend(g.backend, req.Bucket)
```
fsouza, line 47: I realize that this was autoformating, but do you want to take the opportunity to delete this line altogether 😁

fsouza (commented):
Thank you!

internal/grpc/server.go
UnAfraid: Sure thing!

fsouza (approved):
Thank you very much!

UnAfraid:
Thanks for merging it, do you have an idea of when it will be released?

fsouza:
I just pushed a new tag. The release should be available soon!

UnAfraid:
Thank you again!

PR #1519, opened 2024-03-08
Remove unnecessary quoting around ETag in JSON response

### LINKED ISSUE #1327: etag value has extra backslashes when uploading object
When uploading an object against a `fake-gcs-server`, I'm noticing that the `etag` value has escaped backslashes before and after the actual md5 hash. Looks to be due to poor escaping. A GCS client library that I'm using fails to parse the `etag` attribute because of this. I wasn't able to repro when uploading against GCS itself.

Raw json output when uploading against `fake-gcs-server`:
```
"... , \"etag\": \"\\\"1B2M2Y8AsgTpgAmY7PhCfg==\\\"\", ..."
```

Raw json output when uploading against GCS itself:
```
"... , \"etag\": \"TKb9iteyqIEDEEE=\", ..."
```

manuteleco:
The JSON marshaller will already escape the value appropriately. Quoting the value explicitly beforehand is counterproductive, as the quotes introduced in that step would be considered legitimate part of the ETag value and would be themselves escaped when marshalling.

When setting the `ETag` HTTP response header, however, we do have to explicitly add the quoting. This is because double quotes are not added automatically for HTTP headers and, according to the [standard][1], the `ETag` header value _must_ be wrapped in double quotes.

[1]: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/ETag

Fixes #1327

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -854,7 +854,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	w.Header().Set("X-Goog-Generation", strconv.FormatInt(obj.Generation, 10))
 	w.Header().Set("X-Goog-Hash", fmt.Sprintf("crc32c=%s,md5=%s", obj.Crc32c, obj.Md5Hash))
 	w.Header().Set("Last-Modified", obj.Updated.Format(http.TimeFormat))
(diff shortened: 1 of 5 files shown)


fsouza (approved):
Thank you! \o/

PR #1520, opened 2024-03-08
Support the `projection` parameter in `getObject`

manuteleco:
According to the behavior described in the [official documentation][1]. That is, ACL information should only be returned if `projection=full` is provided as GET parameter.

This behavior is not testeable using the Go GCS client, because it hardcodes the value `full` when performing the HTTP request (see [implementation][2]).

[1]: https://cloud.google.com/storage/docs/json_api/v1/objects/get#parameters
[2]: https://github.com/googleapis/google-cloud-go/blob/storage/v1.39.0/storage/http_client.go#L413

### DIFF
--- a/fakestorage/response.go
+++ b/fakestorage/response.go
@@ -9,6 +9,7 @@ import (
 	"net/url"
 	"time"
 
+	"cloud.google.com/go/storage"
 	"github.com/fsouza/fake-gcs-server/internal/backend"
 )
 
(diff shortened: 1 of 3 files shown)


fsouza (commented):
Thank you very much for contributing! This looks good overall.

> This behavior is not testeable using the Go GCS client, because it hardcodes the value full when performing the HTTP request (see [implementation](https://github.com/googleapis/google-cloud-go/blob/storage/v1.39.0/storage/http_client.go#L413)).

Do you think it'd make sense to write a test that sends a low-level request that bypasses the client?

There are some examples in `object_test.go`, such as `TestServerClientObjectUpdateContentType`.

I guess we'd want to test the default behavior and explicitly setting the projection to `noAcl`?

fakestorage/response.go
```diff
@@ -119,6 +120,15 @@ type objectResponse struct {
 	Metadata        map[string]string      `json:"metadata,omitempty"`
 }
 
+func newProjectedObjectResponse(obj ObjectAttrs, projection storage.Projection) objectResponse {
+	objResponse := newObjectResponse(obj)
+	if projection == storage.ProjectionNoACL {
+		objResponse.ACL = nil
+		// TODO: `objResponse.Owner = nil` once we have the `Owner` field in `objectResponse`
```
fsouza, line 127: when resolving the merge conflict, can you also address this TODO?

fsouza (approved):
Thank you very much! This looks good, I can merge it once the merge conflict is resolved.

fakestorage/response.go
manuteleco: I have just removed the TODO instead.

It was probably too speculative to include it to begin with, as it's not really actionable at the moment. Only once (if ever) we implement support for the `owner` attribute in the object response.

fsouza (approved):

PR #1524, opened 2024-03-08
Additional object response attributes

### LINKED ISSUE #758: Self link and media link missing
Hello!

Whenever I use the Google storage java sdk to list blobs in a fake-gcs-server bucket, none of the blobs have their "self link" and "media link" populated. I'm assuming this is a bug?

manuteleco:
Some GCS client libraries (like [this one][1]), are quite strict when it comes to the set of attributes they expect to find in the response. If any of those attributes is missing, they fail to deserialize the response.

Here we introduce three new attributes in the Object response:
* `SelfLink`
* `MediaLink`
* `Metageneration`

`SelfLink` cannot be tested using the Go GCS client library, as it is not included in the [`ObjectAttrs`][2] struct.

For the `Metageneration` we just hardcode the value `"1"`, considering that we are not keeping track of it internally and `"1"` happens to be the correct value most of the times anyway.

[2]: https://pkg.go.dev/cloud.google.com/go/storage@v1.39.0#ObjectAttrs

---

Edited: I believe this PR fixes #758, which I only noticed after the fact.

### DIFF
--- a/fakestorage/object_test.go
+++ b/fakestorage/object_test.go
@@ -15,6 +15,7 @@ import (
 	"hash/crc32"
 	"io"
 	"net/http"
+	"net/url"
 	"reflect"
 	"testing"
 	"time"
@@ -187,7 +188,7 @@ func checkObjectAttrs(testObj Object, attrs *storage.ObjectAttrs, t *testing.T) 
 		}
 	}
 	externalURL := "" // We don't set any `externalURL` value during tests.
-	expectedMediaLink := fmt.Sprintf("%s/download/storage/v1/b/%s/o/%s?alt=media", externalURL, testObj.BucketName, testObj.Name)
+	expectedMediaLink := fmt.Sprintf("%s/download/storage/v1/b/%s/o/%s?alt=media", externalURL, url.PathEscape(testObj.BucketName), url.PathEscape(testObj.Name))
 	if attrs.MediaLink != expectedMediaLink {
 		t.Errorf("wrong MediaLink returned\nwant %s\ngot  %s", expectedMediaLink, attrs.MediaLink)
 	}
--- a/fakestorage/object_test.go
+++ b/fakestorage/object_test.go
@@ -186,8 +186,8 @@ func checkObjectAttrs(testObj Object, attrs *storage.ObjectAttrs, t *testing.T) 
 			t.Errorf("wrong MetaHeader returned\nwant %s\ngot %v", testObj.Metadata["MetaHeader"], val)
 		}
 	}
-	externalUrl := "" // We don't set any `externalURL` value during tests.
-	expectedMediaLink := fmt.Sprintf("%s/download/storage/v1/b/%s/o/%s?alt=media", externalUrl, testObj.BucketName, testObj.Name)
+	externalURL := "" // We don't set any `externalURL` value during tests.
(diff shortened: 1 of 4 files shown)


manuteleco:
Note that this PR will conflict with https://github.com/fsouza/fake-gcs-server/pull/1520 around the definition and use of `newProjectedObjectResponse`. Once one of them is merged, I could rebase and resolve conflicts on the other.

In case that's helpful, here are the changes from all of my PRs from today combined into a single sequence of commits: https://github.com/manuteleco/fake-gcs-server/compare/abe467cf0c6b1c0a55b7afcd31a330f6900241e2...0887d586c15013e5e0f3447c8f976c66839090b4.

fakestorage/object_test.go
```diff
@@ -186,6 +186,14 @@ func checkObjectAttrs(testObj Object, attrs *storage.ObjectAttrs, t *testing.T)
 			t.Errorf("wrong MetaHeader returned\nwant %s\ngot %v", testObj.Metadata["MetaHeader"], val)
 		}
 	}
+	externalUrl := "" // We don't set any `externalURL` value during tests.
```
fsouza, line 189: super nit, but can you call this (and others) `externalURL`?

(I thought there was a vet rule for this in Go, but I guess I'm wrong)

fsouza (commented):
Thank you!

fakestorage/object_test.go
manuteleco: Sure thing. Thanks for pointing it out; I wasn't aware.

fakestorage/response.go
```diff
@@ -140,6 +143,9 @@ func newObjectResponse(obj ObjectAttrs) objectResponse {
 		Updated:         formatTime(obj.Updated),
 		CustomTime:      formatTime(obj.CustomTime),
 		Generation:      obj.Generation,
+		SelfLink:        externalUrl + "/storage/v1/b/" + obj.BucketName + "/o/" + obj.Name,
+		MediaLink:       externalUrl + "/download/storage/v1/b/" + obj.BucketName + "/o/" + obj.Name + "?alt=media",
```
manuteleco, line 147: I should probably URL-encode these. I'll take care of that as well.

fakestorage/object_test.go
manuteleco: Addressed in 97841f3.

fakestorage/response.go
manuteleco: Addressed in 6c65141.

fsouza (approved):
Thank you!

mdedetrich:
@fsouza Is it possible to push a new docker image so we can use this change?, the last one is [2 months old](https://hub.docker.com/r/fsouza/fake-gcs-server/tags) and there are also other new changes which has gotten pushed.

fsouza:
> @fsouza Is it possible to push a new docker image so we can use this change?, the last one is [2 months old](https://hub.docker.com/r/fsouza/fake-gcs-server/tags) and there are also other new changes which has gotten pushed.

Yes, I'll publish a new release later today!

mdedetrich:
> > @fsouza Is it possible to push a new docker image so we can use this change?, the last one is [2 months old](https://hub.docker.com/r/fsouza/fake-gcs-server/tags) and there are also other new changes which has gotten pushed.
> 
> Yes, I'll publish a new release later today!

Many thanks!

PR #1527, opened 2024-03-08
Feature/add example to run vscode.ja

j-antunes:
Hi!

I was trying to run your project with a debugger to test some changes, and I thought it would be useful to provide an example on how to setup a debug session within vscode. 
I find this really helpful to be able to run with a debugger to better follow your code.

This change adds an example `launch.json` file that a dev can use to launch a debug session in vscode.


j-antunes:
@fsouza - That sounds good to me. How can I write to the docs?

PR #1528, opened 2024-03-10
Fill the `errors` attribute for HTTP errors

manuteleco:
Some GCS client libraries (like [this one][1]), are quite strict when it comes to the set of attributes they expect to find in the error response. If any of those attributes is missing, they fail to deserialize the error.

In `fake-gcs-server`, the `httpError.Errors` attribute was always initialized to `nil`, causing the `error.errors` JSON response attribute to be omitted and leading to failures in some client libraries.

In this commit we address that issue by always providing content on the `errors` attribute (`httpError.Errors` field) whenever we return an error response. The implementation is kept simple without aiming to return exactly the same values that are [officially documented][2]. The main goal is to have a sensible response that is _structurally_ correct for the purposes of deserialization client-side.

[2]: https://cloud.google.com/storage/docs/json_api/v1/status-codes

### DIFF
--- a/fakestorage/json_response.go
+++ b/fakestorage/json_response.go
@@ -32,7 +32,7 @@ func jsonToHTTPHandler(h jsonHandler) http.HandlerFunc {
 		status := resp.getStatus()
 		var data any
 		if status > 399 {
-			data = newErrorResponse(status, resp.getErrorMessage(status), nil)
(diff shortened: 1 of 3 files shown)


fsouza (approved):
Thank you for contributing!

PR #1529, opened 2024-03-10
Fix brittle test

manuteleco:
This test was failing sometimes due to a mismatch in clocks between Go and the filesystem, with messages like this one:

```
listed bucket has unexpected creation time. Expected between test
start time 2024-03-10 06:58:36.004471224 +0000 UTC m=+0.048304750
   and now 2024-03-10 06:58:36.006076153 +0000 UTC m=+0.049909699
   actual: 2024-03-10 06:58:35.999266041 +0000 UTC
```


fsouza (approved):
Thank you for contributing and apologies for the delayed review.

PR #1593, opened 2024-05-05
Additional bucket response attributes

manuteleco:
Include these attributes in the JSON response for buckets:

* `projectNumber`
* `metageneration`
* `etag`
* `locationType`

Some [client libraries][1] expect them to be present, and fail to deserialize the response otherwise.

We don't care too much about their values at this stage, so we just picked valid and reasonable constants.


fsouza (approved):
Same as #1594 on hard-coded values. Thanks for contributing!

PR #1594, opened 2024-05-05
Additional ACL response attributes

manuteleco:
Include these attributes in the JSON response for access control lists.

* `etag`
* `kind`

Some [client libraries][1] expect them to be present, and fail to deserialize the response otherwise.

The value for `kind` matches what the real GCS API returns. We don't care too much about the ETag value at this stage, so we just picked a valid constant.


fsouza (approved):
It'd be great to not have hard-coded values, but we can leave that for future contributors!

PR #1608, opened 2024-05-13
read config from environment variables

vichle:
This is a first draft of reading configuration from environment variables (#561) . I chose a different approach than #1041, let me know what you think! I've only added support for a single arg for now: `backend` using the `FAKE_GCS_BACKEND` environment variable. I'll update the PR with the rest of the arguments if you think this is a good way of doing it.

I don't usually write golang, so please be thorough in your feedback :) 

Open questions:
- Is this a good approach?
- Should all arguments be supported?
- Do we need more tests?

### DIFF
--- a/internal/config/config.go
+++ b/internal/config/config.go
@@ -11,6 +11,7 @@ import (
 	"fmt"
 	"log/slog"
 	"math"
+	"os"
 	"strings"
 
 	"github.com/fsouza/fake-gcs-server/fakestorage"
@@ -60,6 +61,13 @@ type EventConfig struct {
 	list            []string
 }
 
+func envVarOrDefault(key string, defaultValue string) string {
+	if val, ok := os.LookupEnv(key); ok {
+		return val
+	}
+	return defaultValue
+}
+
 // Load parses the given arguments list and return a config object (and/or an
 // error in case of failures).
 func Load(args []string) (Config, error) {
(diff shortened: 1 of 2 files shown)


vichle:
@fsouza some feedback would be appreciated :)

fsouza:
> @fsouza some feedback would be appreciated :)

Sorry I completely missed this, will take a look later today!

Isakdl:
@fsouza any love for this? :) Would be great to have support for envs for github actions services.

fsouza (commented):
Apologies for such a bad turnaround.

I think this is fine as it behaves as expected: flag has precedence, but environment variable can be used to override the default value.

If you want to handle all args that'd be appreciated. We need a solution for flags that are not strings, but we don't need to make it too generic given we only have to handle strings and uint.

I'm comfortable with the basic testing you have and maybe exhaustive testing for `envVarOrDefault`.

Thanks for contributing!

ota2000:
First of all, I'd like to express my sincere gratitude to @vichle for initially working on this issue in #1608, and to @fsouza for maintaining this valuable project.

I've taken the initiative to continue and expand upon @vichle's work in #1925, implementing environment variable support for all configuration options. This enhancement is particularly important for containerized environments like GitHub Actions service containers.

I would greatly appreciate your review of #1925, as this functionality would be very beneficial for the community using fake-gcs-server in containerized environments.

Thank you both for your contributions to making fake-gcs-server better!

ota2000:
This has been superseded by #1925, which was released in v1.54.0. Closing this PR.

PR #1696, opened 2024-07-29
add aliases to support specifying EventOptions

### LINKED ISSUE #1695: Unable to specify event notifications when starting using NewServerWithOptions
I wanted to enable event notifications with fake-gcs-server.  It appears that the only way to do so is when using the binary.

[`NewServerWithOptions`](https://pkg.go.dev/github.com/fsouza/fake-gcs-server/fakestorage#NewServerWithOptions) doesn't seem to provide a way to do that.

0xmohit:
This is done to enable specifying EventOptions when using NewServerWithOptions.

https://pkg.go.dev/github.com/fsouza/fake-gcs-server/fakestorage#NewServerWithOptions

Fixes #1695

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -73,6 +73,10 @@ func NewServerWithHostPort(objects []Object, host string, port uint16) (*Server,
 	})
 }
 
+type EventManagerOptions = notification.EventManagerOptions
+
+type EventNotificationOptions = notification.EventNotificationOptions
+
 // Options are used to configure the server on creation.
 type Options struct {
 	InitialObjects []Object
@@ -111,7 +115,7 @@ type Options struct {
 
 	// EventOptions contains the events that should be published and the URL
 	// of the Google cloud function such events should be published to.
-	EventOptions notification.EventManagerOptions
+	EventOptions EventManagerOptions
 
 	// Location used for buckets in the server.
 	BucketsLocation string
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -75,8 +75,6 @@ func NewServerWithHostPort(objects []Object, host string, port uint16) (*Server,
 
 type EventManagerOptions = notification.EventManagerOptions
 
-type EventNotificationOptions = notification.EventNotificationOptions
-
 // Options are used to configure the server on creation.
 type Options struct {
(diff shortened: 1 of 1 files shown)


fakestorage/server.go
```diff
@@ -73,6 +73,10 @@ func NewServerWithHostPort(objects []Object, host string, port uint16) (*Server,
 	})
 }
 
+type EventManagerOptions = notification.EventManagerOptions
+
+type EventNotificationOptions = notification.EventNotificationOptions
```
fsouza, line 78: Can you remove this since it's unused?

fsouza (approved):

fakestorage/server.go
0xmohit: @fsouza I'm not sure if it can be removed.  I was starting the fake server like so:

```go
    return fakestorage.NewServerWithOptions(fakestorage.Options{                                                                                                                                                                                                       
        Scheme:      "http",                                                                                                                                                                                                                                           
        NoListener:  false,                                                                                                                                                                                                                                            
        Host:        host,                                                                                                                                                                                                                                             
        Port:        port,                                                                                                                                                                                                                                             
        PublicHost:  hostPort,                                                                                                                                                                                                                                         
        StorageRoot: storageRoot,                                                                                                                                                                                                                                      
        EventOptions: fakestorage.EventManagerOptions{                                                                                                                                                                                                                 
            ProjectID: projectID,                                                                                                                                                                                                                                      
            TopicName: topicName,                                                                                                                                                                                                                                      
            NotifyOn: fakestorage.EventNotificationOptions{                                                                                                                                                                                                            
                Finalize: true,                                                                                                                                                                                                                                        
                Delete:   true,                                                                                                                                                                                                                                        
            },                                                                                                                                                                                                                                                         
        },                                                                                                                                                                                                                                                             
    })
```

Like `EventManagerOptions`, `EventNotificationOptions` is also defined in an internal package.

fsouza: ohh ok, sorry about that. let's bring it back then. do you want to open a PR or would you rather have me do it? I can push a release right after

0xmohit: I think that you've handled it already.  Thanks :)

PR #1730, opened 2024-08-27
Support setting cache-control of object

gavin726:


gavin726:
Hi, @fsouza , The simulator is very cool! 
I use it in local unit testing. There is a scenario where I will set cache-control of object and need to get and check it. Currently, the simulator does not support it, so this PR supports setting cache-control. Please help review it when you are free, and if there is any problem with it, feel free to ping me, thank you!

.gitignore
```diff
@@ -1 +1,2 @@
 fake-gcs-server
+/.idea
```
fsouza, line 2: I recommend having editor-specific ignore entries set via global gitignore files rather than per-project.

Can you remove this?

fsouza (approved):
Looks good in general. Please update the `.gitignore` file and I can merge it! 😁 

Also, do you not want to set the Cache-Control header in the download object response? (https://github.com/fsouza/fake-gcs-server/blob/5c38ffa16251be9e86b4b3bfd7aa4f6f2565fd25/fakestorage/object.go#L885-L899)

fsouza (approved):

PR #1801, opened 2024-11-05
Disallow uppercase in bucket names

bryfox:
This updates the pattern for valid bucket names to remove support for uppercase characters.

Per https://cloud.google.com/storage/docs/buckets#naming:

> Bucket names can only contain lowercase letters, numeric characters, dashes (-), underscores (_), and dots (.).

I imagine this is a potentially breaking change depending on peoples' setups/fixture data, but this would have caught a bug for one of our projects when changing an internal naming pattern.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -16,7 +16,8 @@ import (
 	"github.com/gorilla/mux"
 )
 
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thanks!

bryfox:
Thanks for making this project!

PR #1813, opened 2024-11-16
Add support for canceling resumable uploads

### LINKED ISSUE #1767: Canceling resumable upload is not supported?
When I tried to [cancel a resumable upload](https://cloud.google.com/storage/docs/performing-resumable-uploads#cancel-upload), the server responded with a 405 Method Not Allowed error.
Is the cancellation of uploads not supported at this time?

Thanks!

kawmra:
Closes #1767 

This PR adds support for canceling resumable uploads.
The implementation is based on the [guides](https://cloud.google.com/storage/docs/performing-resumable-uploads#cancel-upload) and observed behavior, as this functionality was not explicitly documented in the API references.

I'm new to Go, so please feel free to let me know if you have any concerns about naming conventions, code organization, or anything else.

Thanks for maintaining this great project!

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -630,6 +630,10 @@ func parseContentRange(r string) (parsed contentRange, err error) {
 	return parsed, nil
 }
 
+func (s *Server) deleteResumableUpload(r *http.Request) jsonResponse {
+	return jsonResponse{status: 499}
+}
+
 func loadMetadata(rc io.ReadCloser) (*multipartMetadata, error) {
 	defer rc.Close()
 	var m multipartMetadata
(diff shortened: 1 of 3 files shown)


fakestorage/upload.go
```diff
@@ -630,6 +630,10 @@ func parseContentRange(r string) (parsed contentRange, err error) {
 	return parsed, nil
 }
 
+func (s *Server) deleteResumableUpload(r *http.Request) jsonResponse {
+	return jsonResponse{status: 499}
+}
```
kawmra, line 635: Ideally, if a non-existent session URI is specified, the server should return a 404, and a PUT request for a deleted session URI should respond with 499. However, since the upload process did not seem to specifically check whether the session URI actually exists, I decided to follow that approach and not pursue this level of reality.

fsouza: I think that's a fair trade-off.

fsouza (approved):
Thanks for contributing and sorry for the slow review turn around.

PR #1819, opened 2024-11-22
Partial implementation of XML Multipart upload fake

jonmseaman:
Hello, I'm wondering if I can send a partial implementation of the XML Multipart API.

So far, I setup the routing, added the initiate, abort, list uploads, and upload object part apis.
They are barebones implementations and don't support all features listed in the public docs, but it is a good start.

I plan on returning to this when I have time.

Let me know what you think!


fsouza:
@jonmseaman apologies for the delay in getting back to you on this. I like the overall direction here, let me know if you're planning to continue to work on this.

Thank you very much!

jonmseaman:
Thanks for the input

I'll try to get a good v1 done in the next few months

mdedetrich:
Just wanted to say that I am looking forward to the implementation, we use multipart upload heavily as well

jonmseaman:
Thanks for the encouragement! I'll look at it again this week.

jonmseaman:
Ok, I made some updates this past week. You can now upload an entire object and download it with another API. So, the main flow is working now! Let me know what you think. 

There are still a couple improvements I would prefer making, like maybe no having to combine all the parts in-memory to compose the final object, but I think that could be a future change if needed.

jonmseaman:
@fsouza I'm not in a rush or anything, but were you able to take a look at this?

PR #1823, opened 2024-11-25
Persist created and updated time in compose

gaul:
Also use the correct time format.


fsouza (approved):
Thanks for contributing and apologies for the slow turn around.

PR #1824, opened 2024-11-25
Add support for MaxResults when listing objects

gaul:


gaul:
Maybe I missed something but I don't see a way to test this.  Does the Go GCS API lack support for setting MaxResults?

fsouza (approved):
I think it's fine to skip a test here.

PR #1858, opened 2024-12-20
internal/backend: support OpenBSD to get file status last changed info

tuftedocelot:


fsouza (approved):
Looks good. Can you add openbsd as a target in our cross-compile test? https://github.com/fsouza/fake-gcs-server/blob/c7ac293e3ec6757b28f151cee34d666e733973c9/.github/workflows/main.yml#L12-L26

tuftedocelot:
Like so?

fsouza (approved):
Thank you!

PR #1902, opened 2025-02-14
Use default temporary directory in tests

commiterate:
Removing the Linux temporary directory hardcode in unit tests as it breaks sandboxed builds.

---

For packaging `fake-gcs-server` in Nixpkgs (will need a new release with the fixes).

### DIFF
--- a/fakestorage/bucket_test.go
+++ b/fakestorage/bucket_test.go
@@ -8,7 +8,6 @@ import (
 	"context"
 	"errors"
 	"os"
-	"runtime"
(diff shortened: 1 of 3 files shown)


fsouza:
fake-gcs-server only has one command, where the convention is to keep main on the root (it's documented in the page you linked).

Happy to merge the temp dir change though, provided it doesn't break tests in other OSes (GitHub Actions will check that for us).

fsouza (approved):
Thanks for contributing!

PR #1970, opened 2025-05-26
fakeserver: honor Stop

motiejus:
Problem: calling `server.Close()` does not actually close anything — requests can flow as they normally would.

That's because the client always uses `*muxTransport`, which wires directly into the handlers avoiding all the TCP/HTTP dance.

In days long gone, `Server.transport` used to be HTTP, and starting/stopping the fakeserver did what one would expect.

Add a test that asserts correct behavior (we depend on it in some our tests), also removing the `Server.transport` ambiguity, as there is no other use case in this repo.

### DIFF
--- a/fakestorage/mux_transport.go
+++ b/fakestorage/mux_transport.go
@@ -5,15 +5,20 @@
 package fakestorage
 
 import (
+	"errors"
 	"net/http"
 	"net/http/httptest"
 )
 
 type muxTransport struct {
 	handler http.Handler
+	closed  bool
 }
 
 func (t *muxTransport) RoundTrip(r *http.Request) (*http.Response, error) {
+	if t.closed {
+		return nil, errors.New("server closed")
+	}
 	w := httptest.NewRecorder()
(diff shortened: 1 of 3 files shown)


motiejus:
Having made this PR, I am not entirely sure what's the role of the explicit listeners, given that the client is always created directly attached to `s.transport`, which points direclty to `*muxHandler`?

https://github.com/fsouza/fake-gcs-server/blob/79d71622f893b4661a435b34ca9a197a911a8d43/fakestorage/server.go#L493-L496

fsouza:
> Having made this PR, I am not entirely sure what's the role of the explicit listeners, given that the client is always created directly attached to s.transport, which points direclty to *muxHandler?

The code needs some cleanup, but the listeners are used when fake-gcs-server runs as a standalone binary. If you're using it as a library from Go, then you only interact with muxTransport. Your change is correct

motiejus:
So which one is it? :)

> We need to keep it as RoundTripper

> Your change is correct

Happy to change it back to `http.RoundTripper`, let me know.

PR #1979, opened 2025-06-13
Fix direct dependency on grpc/stats/opentelemetry to avoid conflicts

DenKoren:
The dependency on `google.golang.org/grpc/stats/opentelemetry` instead of `google.golang.org/grpc` package makes it impossible to incorporate `fake-gcs-server/fakestorage` into go tests that use modern version of gcloud SDK for go.
Even module module replacements for `google.golang.org/grpc/stats/opentelemetry` do not solve the problem.
The only way is to replace `fake-gcs-server/fakestorage` with custom fork.

This version update followed by `go mod tidy` solves the problem.


fsouza (approved):
Thanks for the fix!

I wonder why dependabot didn't upgrade it, but I'm glad you fixed it

PR #1985, opened 2025-07-07
Add support for MatchGlob in list queries

### LINKED ISSUE #1415: Server does not support MatchGlob for list Queries
[google cloud storage package](https://pkg.go.dev/cloud.google.com/go/storage#Query) provides a nifty way to select objects using the [glob expressions](https://cloud.google.com/storage/docs/json_api/v1/objects/list#list-object-glob) . 

```go
type Query struct {
        ...
	// MatchGlob is a glob pattern used to filter results (for example, foo*bar). See
	// https://cloud.google.com/storage/docs/json_api/v1/objects/list#list-object-glob
	// for syntax details. When Delimiter is set in conjunction with MatchGlob,
	// it must be set to /.
	MatchGlob [string](https://pkg.go.dev/builtin#string)
	// contains filtered or unexported fields
}
```

Unfortunately, fake server doesn't support this kind of queries.

I've verified by adding couple more test cases for listing objects which are actually failing ([0001-add-tests-for-MatchGlob-failing.patch](https://github.com/fsouza/fake-gcs-server/files/13587136/0001-add-tests-for-MatchGlob-failing.patch) )

```diff
diff --git a/fakestorage/object_test.go b/fakestorage/object_test.go
index b158158..4ac9660 100644
--- a/fakestorage/object_test.go
+++ b/fakestorage/object_test.go
@@ -916,6 +916,30 @@ func getTestCasesForListTests(versioningEnabled, withOverwrites bool) []listTest
                        },
                        nil,
                },
+               {
+                       fmt.Sprintf("filtering MatchGlob *, versioning %t and overwrites %t", versioningEnabled, withOverwrites),
+                       "some-bucket",
+                       &storage.Query{MatchGlob: "img/*.jpg"},
+                       []string{
+                               "img/brand.jpg",
+                       },
+                       nil,
+               },
+               {
+                       fmt.Sprintf("filtering MatchGlob **, versioning %t and overwrites %t", versioningEnabled, withOverwrites),
+                       "some-bucket",
+                       &storage.Query{MatchGlob: "img/**.jpg"},
+                       []string{
+                               "img/brand.jpg",
+                               "img/hi-res/party-01.jpg",
+                               "img/hi-res/party-02.jpg",
+                               "img/hi-res/party-03.jpg",
+                               "img/low-res/party-01.jpg",
+                               "img/low-res/party-02.jpg",
+                               "img/low-res/party-03.jpg",
+                       },
+                       nil,
+               },
                {
                        fmt.Sprintf("full prefix, versioning %t and overwrites %t", versioningEnabled, withOverwrites),
                        "some-bucket"
```
So it would be nice if this functionality could be added in further releases

sergeyevstifeev:
Fixes https://github.com/fsouza/fake-gcs-server/issues/1415


pinguet62:
Hello 🖖
Any news ? 👼

fakestorage/object.go
```diff
@@ -1277,3 +1286,120 @@ func (s *Server) composeObject(r *http.Request) jsonResponse {
...
+// - ** matches zero or more characters (including /)
+// - ? matches any single character (excluding /)
+// - [abc] matches exactly one character in the set
+// - [a-z] matches one character in a range
+// - [!abc] or [^abc] matches one character NOT in the set
+// - {abc,xyz} matches one of the specified options
+// Doesn't currently support:
+// - nested {} braces
+// - patterns inside {} braces e.g. {*foo,bar*}
+func matchGlob(pattern, name string) bool {
```
fsouza: can we use doublestar instead of the custom globToRegex? Basically we can replace all calls to `matchGlob` to calls to `doublestar.MatchUnvalidated`

fakestorage/object_test.go
```diff
@@ -986,6 +986,117 @@ func getTestCasesForListTests(versioningEnabled, withOverwrites bool) []listTest
...
+			[]string{
+				"img/hi-res/party-01.jpg",
+				"img/hi-res/party-02.jpg",
+				"img/low-res/party-01.jpg",
+				"img/low-res/party-02.jpg",
+			},
+			nil,
+		},
+		{
+			fmt.Sprintf("filtering MatchGlob ** and \a, versioning %t and overwrites %t", versioningEnabled, withOverwrites),
```
fsouza:
```suggestion
			fmt.Sprintf(`filtering MatchGlob ** and \a, versioning %t and overwrites %t`, versioningEnabled, withOverwrites),
```

fsouza (commented):
Thanks for contributing, and apologies for the delayed review

zachvictor:
This PR looked stalled after the review in January, so I opened #2328 to carry it forward. It keeps @sergeyevstifeev's commit as-is and adds one commit that addresses both review threads. Happy to close mine if work resumes here instead.

PR #1989, opened 2025-07-17
Add support for NetBSD

iamleot:
When trying to build fake-gcs-server on NetBSD it fails with:

```
# github.com/fsouza/fake-gcs-server/internal/backend
internal/backend/fs.go:120:82: undefined: createTimeFromFileInfo
internal/backend/fs.go:156:82: undefined: createTimeFromFileInfo
```

This PR addresses that and now fake-gcs-server builds on NetBSD too.

(I have noticed that while updating Grafana Loki.)

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -20,6 +20,7 @@ jobs:
           - darwin
           - freebsd
           - linux
+          - netbsd
           - openbsd
           - windows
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thanks for contributing!

iamleot:
Thank you!

PR #2063, opened 2025-11-06
feat: Added support for "X-Goog-Stored-Content-Length" in object download headers.

mmukada:
According to the Google team, the `X-Goog-Stored-Content-Length` header is always included in GCS object downloads and corresponds to `metadata.size` (the object’s size at rest). The Rust SDK uses this header to convey the object size during downloads — see [https://github.com/googleapis/google-cloud-rust/discussions/3713](https://github.com/googleapis/google-cloud-rust/discussions/3713).

Documentation: [https://docs.cloud.google.com/storage/docs/json_api/v1/parameters#xgoogstoredcontentlength](https://docs.cloud.google.com/storage/docs/json_api/v1/parameters#xgoogstoredcontentlength)

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -1013,6 +1013,7 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 	lastByte := int64(0)
 	satisfiable := true
 	contentLength := int64(0)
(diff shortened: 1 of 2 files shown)


fsouza (dismissed):
Thank you!

Issue #1933, opened 2025-03-18
Fake GCS adds .metadata file to *.metadata files


AlvesJorge:
Hi there,
First of all, regular user of fake-gcs here, thank you for such a great service :) 

The issue: It seems that fake gcs create metadata files for the metadata files it creates, which means every couple of weeks I have to run `for f in **/**.metadata.metadata; do rm "$f"; done` because the container eventually can't handle all those files. 

Not ruling out a configuration / app error on my part though, but I still want to put it out there in case a contributor sees this and it problem "clicks" with them or someone else is experiencing this.

sam-njia:
Also experiencing this, and so are my colleagues.

I may work on a fix sometime

<img width="2074" height="922" alt="Image" src="https://github.com/user-attachments/assets/46c390bf-ef0f-420f-b0aa-7bc57bcc264f" />

francisco-barros-codes:
Still a relevant issue

PR #2064, opened 2025-11-06
Prevent creating metadata files of metadata

usr-ein:
Relating to issue #1933 

Idk if this solves the issue, still need to try it out, but I reckon that's the cleanest way to do this.

### DIFF
--- a/fakestorage/server.go
+++ b/fakestorage/server.go
@@ -410,6 +410,12 @@ func objectsFromBucket(localBucketPath, bucketName string) ([]Object, error) {
 	var objects []Object
 	err := filepath.Walk(localBucketPath, func(path string, info os.FileInfo, _ error) error {
 		if info.Mode().IsRegular() {
+			// Skip internal metadata files so they don't get re-ingested as
+			// objects on restart, which would cause recursive
+			// .metadata.metadata.metadata... chains (see #1933).
+			if strings.HasSuffix(info.Name(), ".metadata") {
+				return nil
+			}
 			// Rel() should never return error since path always descend from localBucketPath
 			relPath, _ := filepath.Rel(localBucketPath, path)
 			objectKey := filepath.ToSlash(relPath)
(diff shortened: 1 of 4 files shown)


aman-vizibl:
Any update on this? I have been facing the same issue

fsouza:
hmmm I think the correct fix will require a big change to the fs backend, no? Like we'll need to store metadata separately?

usr-ein:
@fsouza Ready for review, I've made a few more changes that definitely take care of this:

- First one fixes the actual root cause — `objectsFromBucket()` was walking the seed directory and picking up `.metadata` files as if they were regular objects, so on every restart `CreateObject` would write a new layer of `.metadata.metadata.metadata....` Just skipping files ending in `.metadata` during the walk, same as `ListObjects()` already does.

- Second one adds a guard in `CreateObject` that errors out if the object name ends in `.metadata` when using the file-based metadata handler. Probably redundant with the first fix but I'd rather have it there.

- Third one cleans up the mess on existing deployments — on init it walks the storage dir and deletes anything with chained `.metadata.metadata` in the name, since those are 100% garbage from the bug. Logs a warning so you see it happened.

Tests for all three, full suite passes.

usr-ein:
Any update on this? It's been long enough that I'm likely to run in the bug again after the number of restarts from the container I've had ._.

fsouza:
Thanks for sticking with this, and sorry for the slow turnaround.

Commits 1 + 5 look right. I reproduced #1933 locally (seed dir == filesystem root, forcing the file-based metadata handler, since xattr masks this on ext4/APFS): `main` grows one `.metadata` layer per restart, this branch stays flat.

Could you drop commit 3 (the startup cleanup)? Two problems:

- `strings.Contains(info.Name(), ".metadata.metadata")` matches anywhere in the name, not just as a suffix. An object legitimately named `report.metadata.metadata.json` gets deleted along with its sidecar on the next start — I hit this with an object created through the API, no seeding involved.
- It conflicts with commit 5: `notes.metadata.metadata` is the real sidecar for the `notes.metadata` object that commit 5 deliberately keeps, and the cleanup removes it.

More generally, I'd rather not delete files from a user's data dir on every boot without an explicit opt-in. Once commit 1 is in no new chains are created, and the existing ones are inert leftovers people can remove themselves.

With commit 3 dropped and a rebase on main, this is good to go.


🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2103, opened 2026-01-03
Fix metadata headers not being preserved during upload and retrieval

### LINKED ISSUE #2102: Object metadata not preserved (Cache-Control, Content-Encoding, Content-Disposition, custom metadata)
Object metadata set during upload is lost and returns `null` when retrieving objects. This happens for both simple uploads and multipart uploads.

## What's broken

When uploading with metadata like:

```java
BlobInfo.newBuilder(bucket, name)
    .setCacheControl("max-age=3600")
    .setContentEncoding("gzip")
    .setContentDisposition("attachment; filename=doc.pdf")
    .setMetadata(Map.of("owner", "user123"))
    .build();
```

All these values come back as `null` on subsequent GET requests.

## Why this matters for us

We're using fake-gcs-server to test [S3Proxy](https://github.com/gaul/s3proxy)'s new native GCS SDK backend (`google-cloud-storage-sdk`). S3Proxy translates S3 API calls to GCS, so we need metadata to round-trip correctly.

Without this, we can't verify that:

- S3 `Cache-Control` headers survive the S3 -> GCS -> S3 translation
- Multipart uploads preserve the metadata set during `InitiateMultipartUpload`
- `Content-Disposition` for downloads works correctly
- Custom S3 user metadata (`x-amz-meta-*`) maps to GCS metadata properly

We've had to skip these assertions in our test suite when running against fake-gcs-server, which means less confidence in local testing before hitting real GCS.

## Tested with

- `fsouza/fake-gcs-server:latest` with `-scheme http` on port 4443
- Google Cloud Storage Java SDK 2.x.x

All these tests pass against real GCS, so the issue is specific to fake-gcs-server.

louisnow:
## Summary

Fixes #2102 - Object metadata like Cache-Control, Content-Disposition, Content-Language, and custom `x-goog-meta-*` headers were being silently dropped during uploads.

The problem was that several upload handlers weren't reading these headers from the request or passing them through to the stored object. Additionally, Cache-Control wasn't being returned in JSON API responses, the gRPC server was missing metadata fields, and ComposeObject wasn't preserving encoding/cache settings.

## Changes

**JSON API response fixes:**
- Add `cacheControl` field to `objectResponse` struct so it gets returned to clients
- Add `cacheControl` to `jsonObject` struct and its JSON marshal/unmarshal methods
- Add `cacheControl` to `bufferedObjectsToBackendObjects` conversion
- Fix JSON tag typo in `multipartMetadata` struct (`"ContentLanguage"` → `"contentLanguage"`)

**Upload handler fixes:**
- Simple uploads: add Content-Disposition, Content-Language, and custom `x-goog-meta-*` header support
- Signed URL uploads: add Cache-Control, Content-Disposition, Content-Language support
- Resumable uploads: add Content-Disposition, Content-Language support
- Form/multipart uploads: add Cache-Control, Content-Disposition, Content-Language support
- Multipart uploads: fix contentEncoding query parameter handling

**Object operation fixes:**
- Add CacheControl to patchObject and updateObject handlers
- Preserve Cache-Control during object rewrite/copy operations
- Add contentEncoding and cacheControl parameters to ComposeObject interface and implementations (memory + filesystem backends)

**gRPC server fixes:**
- Fix GetObject to include all metadata fields
- Fix UpdateObject to include all metadata fields
- Fix PatchObject to include CacheControl
- Fix ListObjects to include all metadata fields
- Fix makeObject helper to include all metadata fields

**Tests:**
- Add HTTP integration tests for simple uploads, signed uploads, and form uploads with all metadata headers
- Add test for contentEncoding query param in multipart uploads
- Add test for patching CacheControl on existing objects
- Update gRPC tests to verify all metadata fields (ContentEncoding, ContentDisposition, ContentLanguage, CacheControl, Metadata)
- Add Cache-Control and Content-Language checks to existing object writer and rewrite tests

### DIFF
--- a/fakestorage/upload_test.go
+++ b/fakestorage/upload_test.go
@@ -643,7 +643,6 @@ func TestServerClientMultipartUploadWithContentEncodingQueryParam(t *testing.T) 
 	if err != nil {
 		t.Fatal(err)
 	}
-	req.Header.Set("Content-Type", writer.FormDataContentType())
 	// Fix the content type to be multipart/related instead of multipart/form-data
 	req.Header.Set("Content-Type", "multipart/related; boundary="+writer.Boundary())
 


fsouza:
@louisnow thanks for contributing! can you rebase against the latest main? I've made some changes to ci

louisnow:
@fsouza I've rebased the changes and tested it, all tests pass as well.

fakestorage/upload_test.go
```diff
@@ -531,6 +536,150 @@ func TestServerClientSimpleUpload(t *testing.T) {
...
+	}
+
+	writer.Close()
+
+	// Send request with contentEncoding as query parameter (not in metadata JSON)
+	req, err := http.NewRequest("POST", server.URL()+"/upload/storage/v1/b/other-bucket/o?uploadType=multipart&contentEncoding="+contentEncoding, &buf)
+	if err != nil {
+		t.Fatal(err)
+	}
+	req.Header.Set("Content-Type", writer.FormDataContentType())
```
fsouza: nit: can you remove this line? we're overriding the header right after anyways

PR #2114, opened 2026-01-21
fix: Preserve content type from resumable upload session metadata

ide:
Why
===
When uploading files using resumable uploads, the `contentType` specified in the initial session creation request was being overwritten with `application/octet-stream` when uploading file content.

Before: session metadata sets `contentType: "text/csv"` → chunk upload without Content-Type header → object has `contentType: "application/octet-stream"`

After: session metadata sets `contentType: "text/csv"` → chunk upload without Content-Type header → object has `contentType: "text/csv"`

How
===
Only default to `application/octet-stream` when no content type was set in either the chunk upload's Content-Type header or the session initialization metadata.

Test Plan
===
`go test ./fakestorage/... -run TestResumableUploadContentType`

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -522,7 +522,7 @@ func (s *Server) uploadFileContent(r *http.Request) jsonResponse {
 	contentTypeHeader := r.Header.Get(contentTypeHeader)
 	if contentTypeHeader != "" {
 		obj.ContentType = contentTypeHeader
-	} else {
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thank you for contributing!

fsouza:
@ide can you rebase against main to pickup the fix?

ide:
Thanks, I rebased on top of main.

PR #2117, opened 2026-01-23
Implement per-object locks for better parallelism in filesystem backend

### LINKED ISSUE #1978: support parallel uploads
When a client attempts parallel uploads of new objects (e.g. hundreds of unique small files) the upload latency quickly climbs to several seconds.

The backends have a global [RWMutex](https://github.com/fsouza/fake-gcs-server/blob/main/internal/backend/memory.go#L24). Per-object locks could reduce contention.

environment: fake-gcs-server and single client deployed in local minikube-on-docker

misha-drozd:
Fixes #1978 for filesystem backend

### DIFF
--- a/internal/backend/fs.go
+++ b/internal/backend/fs.go
@@ -196,7 +196,10 @@ func getBucketAttributes(path string) (BucketAttrs, error) {
 
 // DeleteBucket removes the bucket from the backend.
 func (s *storageFS) DeleteBucket(name string) error {
-	objs, err := s.ListObjects(name, "", false)
+	s.mtx.Lock()
+	defer s.mtx.Unlock()
+
+	objs, err := s.listObjects(name, "", false)
 	if err != nil {
 		return BucketNotFound
 	}
@@ -204,8 +207,6 @@ func (s *storageFS) DeleteBucket(name string) error {
 		return BucketNotEmpty
 	}
 
-	s.mtx.Lock()
-	defer s.mtx.Unlock()
 	return os.RemoveAll(filepath.Join(s.rootDir, url.PathEscape(name)))
 }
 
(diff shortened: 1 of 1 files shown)


internal/backend/fs.go
```diff
@@ -381,32 +423,49 @@ func (s *storageFS) getObjectAttrs(bucketName, objectName string) (ObjectAttrs,
...
 	if err := s.mh.remove(path); err != nil {
 		return err
 	}
-	return os.Remove(path)
+	if err := os.Remove(path); err != nil {
+		return err
+	}
+
+	// Clean up the lock entry
+	s.deleteObjectLock(bucketName, objectName)
```
fsouza: this is racy. we either need to version the lock or not delete object locks.

(alternatively we accept that delete will be slow and simply keep the previous version that locked the bucket)

fsouza (commented):
ComposeObject still has a race condition. It reads source objects one by one (acquiring and releasing each read lock), then creates the destination. Between reading sources and creating the destination, another goroutine could delete/modify a source object.

```go
func (s *storageFS) ComposeObject(...) (StreamingObject, error) {
    for _, n := range objectNames {
        obj, err := s.GetObject(bucketName, n)  // acquires & releases read lock
        // ...
    }
    // No locks held here! Source objects could be modified/deleted
    
    result, err := s.CreateObject(dest, NoConditions{})  // acquires dest write lock
}
```

To fix this properly, ComposeObject would need to acquire read locks on all source objects AND a write lock on the destination, in a consistent order (e.g., sorted by name) to avoid deadlocks, and hold them all until the operation completes.

Alternatively, ComposeObject could keep using the global lock for simplicity since it's a less common operation.

misha-drozd:
> To fix this properly, ComposeObject would need to acquire read locks on all source objects AND a write lock on the destination, in a consistent order (e.g., sorted by name) to avoid deadlocks, and hold them all until the operation completes.

Fixed.

I had a look around more and fixed a small bug in `DeleteBucket`: `os.RemoveAll` might fail if another object is being created while it is running (as it does "listdir" + "foreach(file) rm" inside - here's a tiny race window).

Another potential issue I left "as is" for now: `GetObjectWithGeneration` might be racy if between `GetObject` and reading the body the object is updated: as file handle is opened lazily, there's TOCTOU mismatch.

misha-drozd:
Fixed lint issue.

misha-drozd:
> Any thoughts on how we can fix this?

Maybe eagerly walk and open the file handles while ListObjects is holding the lock? I have already been bitten once by lazy file handle opening earlier in this change.

PR #2252, opened 2026-06-26
fix: return 400 for negative maxResults instead of panicking

gaul:
A negative maxResults query parameter (e.g. ?maxResults=-1) parsed successfully via strconv.Atoi and flowed into the pagination logic, where respObjects[options.MaxResults] panicked with index out of range. Real GCS treats maxResults as an unsigned integer and rejects invalid values with 400 Bad Request.

Reject negative values at the handler with 400, and guard the pagination slice with MaxResults > 0 so direct callers of the exported ListObjectsWithOptionsPaginated cannot panic the process either.


fsouza (approved):
Thank you!

PR #2253, opened 2026-06-26
fix: return Content-Range and full body on 416 range responses

gaul:
When a Range request's start is beyond the object size, downloadObject returned 416 but set Content-Length: 0 (computed from the unsatisfiable range), which truncated the XML error body to empty, and it never set a Content-Range header.

Real GCS answers an unsatisfiable range with 'Content-Range: bytes */<size>' and an InvalidRange error body. Set that header and override Content-Length with the body length so the error body is actually sent.


fsouza (approved):
Thank you!

PR #2255, opened 2026-06-26
fix: return 200 on POST request to pre-signed URL made by gcloud CLI

### LINKED ISSUE #2254: 400 Bad Request after POST using pre-signed URL generated by gcloud CLI
# Problem
When performing a POST request on a pre-signed URL to initiate a resumable upload, the server responds with `400 Bad Request` (an equivalent request made to GCS is successful).

To reproduce, generate a pre-signed URL using the gcloud CLI:
```zsh
gcloud storage sign-url gs://test-bucket/test-object --http-verb=POST --headers=x-goog-resumable=start,content-type=application/octet-stream --impersonate-service-account=[ACCOUNT] --region=us-east4
```

Then POST using the resulting URL:
```zsh
curl -i -X POST "http://test-bucket.localhost:4443/test-object?x-goog-signature=[SIGNATURE]&x-goog-algorithm=GOOG4-RSA-SHA256&x-goog-credential=[CREDENTIAL]&x-goog-date=[DATE]&x-goog-expires=3600&x-goog-signedheaders=content-type%3Bhost%3Bx-goog-resumable" \
  -H "x-goog-resumable: start" \
  -H "Content-Type: application/octet-stream"
```

Response is
```
HTTP/1.1 400 Bad Request
Content-Type: application/json
Date: Fri, 26 Jun 2026 14:13:27 GMT
Content-Length: 139

{"error":{"code":400,"message":"invalid uploadType","errors":[{"domain":"global","reason":"Bad Request","message":"invalid uploadType"}]}}
```

# Expected Behavior
The server should return `200 OK` or `201 Created` along with a sessionURI, mimiking the behavior of the actual API: [resumable upload response example](https://docs.cloud.google.com/storage/docs/xml-api/post-object-resumable#example)

i.e. follow the same procedure as above an instead receive the response:
```
HTTP/1.1 200 OK
Content-Type: application/json
Location: http://test-bucket.localhost:4443/upload/storage/v1/b/test-bucket/o?uploadType=resumable&name=&upload_id=ee0eafdcd11110ea4d206ef84bcb23d1
Date: Fri, 26 Jun 2026 14:21:55 GMT
Content-Length: 481

{"kind":"storage#object","name":"","id":"test-bucket/","bucket":"test-bucket","size":"0","acl":[{"bucket":"test-bucket","entity":"projectOwner-test-project","etag":"RVRhZw==","kind":"storage#objectAccessControl","projectTeam":{},"role":"OWNER"}],"storageClass":"STANDARD","generation":"0","selfLink":"http://test-bucket.localhost:4443/storage/v1/b/test-bucket/o/","mediaLink":"http://test-bucket.localhost:4443/download/storage/v1/b/test-bucket/o/?alt=media","metageneration":"1"}
```

samneisewander:
# Problem
The current code checks for the presence of the `X-Goog-Algorithm` query parameter in order to dispatch this kind of request to the `(s *Server) resumableUpload` slot function. However, the capitalization of this query parameter depends on the way the pre-signed URL was created. Server applications that create signed URLs are instructed to capitalize this query parameter per the documentation: [signing URLs manually](https://docs.cloud.google.com/storage/docs/access-control/signing-urls-manually#algorithm)

However, signed URLs created using the gcloud CLI make all of the required query parameters lowercase:
```zsh
gcloud storage sign-url gs://test-bucket/test-object --http-verb=POST --headers=x-goog-resumable=start,content-type=application/octet-stream --impersonate-service-account=[ACCOUNT] --region=us-east4

# outputs URL with "x-goog-signature", "x-goog-algorithm", etc
```

This is interesting behavior because the required query string parameters in a [canonical request](https://docs.cloud.google.com/storage/docs/authentication/canonical-requests#required-query-parameters) are pascal cased... (could it be a bug in the gcloud cli?)

# Solution
All POST requests using pre-signed URLs that initiate a resumable upload using the XML API have a query parameter `x-goog-signedheaders` or `X-Goog-SignedHeaders` containing a `;` delimited list of names of headers defined in the canonical request. Resumable upload requests must have the header `x-goog-resumable: start`, so `x-goog-resumable` will appear in this list if this URL is for a resumable upload:
- https://docs.cloud.google.com/storage/docs/authentication/canonical-requests#required-query-parameters
- https://docs.cloud.google.com/storage/docs/performing-resumable-uploads#initiate-session

The patch just checks if this query parameter exists on the URL and contains `x-goog-resumable`. If so, it dispatches to the `(s *Server) resumableUpload` slot function.

Fixes  #2254

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -639,6 +639,10 @@ func (s *Server) resumableUpload(bucketName string, r *http.Request) jsonRespons
 	if objName == "" {
 		objName = metadata.Name
 	}
+	// For XML resumable upload, object name is the path. Omit leading slash.
+	if objName == "" {
+		objName = strings.TrimPrefix(r.URL.Path, "/")
+	}
 	if contentEncoding == "" {
 		contentEncoding = metadata.ContentEncoding
 	}
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -16,6 +16,7 @@ import (
 	"net/http"
 	"net/url"
 	"regexp"
+	"slices"
 	"strconv"
 	"strings"
 	"time"
@@ -159,6 +160,19 @@ func (s *Server) insertObject(r *http.Request) jsonResponse {
 				return s.signedUpload(bucketName, r)
 			}
 		}
+
+		// Support initiating resumable upload using XML API
+		var headers []string
+		switch {
+		case r.URL.Query().Has("X-Goog-SignedHeaders"):
+			headers = strings.Split(r.URL.Query().Get("X-Goog-SignedHeaders"), ";")
+		case r.URL.Query().Has("x-goog-signedheaders"):
+			headers = strings.Split(r.URL.Query().Get("x-goog-signedheaders"), ";")
+		}
+		if slices.Contains(headers, "x-goog-resumable") {
+			return s.resumableUpload(bucketName, r)
+		}
+
 		return jsonResponse{errorMessage: "invalid uploadType", status: http.StatusBadRequest}
 	}
 }


samneisewander:
Btw, this is literally my first open source contribution ever so please look with extra scrutiny!! Feedback is appreciated.

fsouza (commented):
Thanks for the detailed writeup — the diagnosis here is excellent, and the gcloud casing discrepancy is a real bug worth fixing. I dug into it and have two concerns, both verified by running the code on this branch.

## 🔴 Object name includes the bucket for path-style URLs

`objName = strings.TrimPrefix(r.URL.Path, "/")` assumes the path is *only* the object name. That holds for virtual-host style (`/{objectName:.+}`, `server.go:336-337`) but **not** for path-style, which routes via `/{bucketName}/{objectName:.+}` (`server.go:333`). Path-style is exactly what `gcloud storage sign-url gs://test-bucket/test-object` produces — i.e. the scenario this PR targets.

Reproduced against this branch:

```
POST /test-bucket/test-object?x-goog-algorithm=...&x-goog-signedheaders=content-type;host;x-goog-resumable
→ Location: ...&name=test-bucket%2Ftest-object&upload_id=...   # want name=test-object
```

So the object lands at `test-bucket/test-object` inside `test-bucket`. The sibling `signedUpload` already gets this right at `upload.go:463` — using the route var works:

```go
if objName == "" {
    objName = unescapeMuxVars(mux.Vars(r))["objectName"]
}
```

I verified this yields `name=test-object`. It also handles percent-decoding correctly, which `TrimPrefix` on `r.URL.Path` does not.

## 🟡 This treats the symptom rather than the cause — PUT is still broken

The root cause is the case-sensitive parameter lookup, not missing resumable detection. Since `POST` + algorithm already dispatches to `resumableUpload`, the entire `SignedHeaders` block turns out to be unnecessary — and because it's POST-only, a gcloud-signed **PUT** still fails:

```
PUT /test-bucket/put-object?x-goog-algorithm=... → 400 invalid uploadType
```

Matching the existing param case-insensitively fixes both verbs and removes more code than it adds:

```go
// Query parameter casing varies by signer (gcloud CLI emits lowercase).
if hasSignedURLParam(r.URL.Query(), "x-goog-algorithm") {
    switch r.Method {
    case http.MethodPost:
        return s.resumableUpload(bucketName, r)
    case http.MethodPut:
        return s.signedUpload(bucketName, r)
    }
}
```

...with a small helper using `strings.EqualFold` over the query keys (which also lets the `slices` import go away). I ran this locally: the gcloud POST **and** PUT both return 200, and `go test -race -vet all ./fakestorage` stays green. It also naturally covers any other casing a signer might emit, rather than the two hardcoded cases.

## 🟡 Tests

There are no tests here, and both issues above would have been caught by one. `TestServerClientSignedUpload` (`upload_test.go:682`) is a ready-made template — worth adding path-style and vhost-style cases with a gcloud-style lowercase URL, asserting the resulting object name.

One gotcha that'll save you time: the test server needs `PublicHost: "127.0.0.1"`, otherwise `publicHostMatcher` won't match and you'll get a misleading 404 rather than the code path you're aiming at.

## Minor

- `r.URL.Query()` is re-parsed several times in the new block; parse once into a local. Moot if you take the case-insensitive approach.

## Summary

Real issue, and I'd like to get a fix in. My suggestion is to collapse this into the case-insensitive algorithm check plus the `mux.Vars` object name, with a test — that ends up a smaller diff than the current one while fixing strictly more. The object-name bug is the blocker as-is: it means the reported scenario still misbehaves silently, which is arguably worse than the 400 it replaces.

samneisewander:
Sounds good, I appreciate the thorough review! I will work on addressing your concerns in this PR. Do you have a preferred workflow for making changes and then asking for another review?

fsouza:
> Sounds good, I appreciate the thorough review! I will work on addressing your concerns in this PR. Do you have a preferred workflow for making changes and then asking for another review?

If you can make additional commits instead of amending/squashing, then it's easier for me to review the diff in github! This is a relatively small diff though, so feel free to ignore that lol

PR #2256, opened 2026-06-26
fix: prevent gRPC ComposeObject panic with more than two sources

gaul:
The gRPC ComposeObject handler allocated a fixed-length slice of 2 for the source object names, so composing more than two sources wrote past the end of the slice and panicked with "index out of range [2] with length 2". Fewer than two sources left trailing empty names. Allocate the slice to match the number of source objects.


fsouza (approved):
Thanks for the fix and for including a test!

PR #2257, opened 2026-06-26
fix: use GCS reason enums in JSON error responses

gaul:
The JSON API error model returns a domain-specific reason enum, but the errors[].reason field was populated with the HTTP status text, e.g. "Not Found" instead of "notFound" and "Precondition Failed" instead of "conditionNotMet". Map the status codes the server emits to their GCS reason strings.

Two existing tests asserted the previous, incorrect reason text inside full error strings; update them to the real reasons (conditionNotMet, invalid).


fsouza (approved):
Thanks for the fix!

PR #2260, opened 2026-06-27
fix: enforce generation preconditions on resumable uploads

gaul:
Resumable uploads accepted ifGenerationMatch/ifGenerationNotMatch on the session-initiating request but discarded them when finalizing the upload, so the preconditions were silently ignored. Only the single-request multipart path enforced them.

GCS SDKs (e.g. the Java client's Storage.createFrom) use resumable uploads, so an If-None-Match: * (ifGenerationMatch=0) overwrite of an existing object incorrectly returned 200 instead of 412.

Carry the preconditions across the two-step flow: capture them with the upload session at initiation (resumableUpload and
handleBodyBasedResumableUpload) and enforce them when the object is created in uploadFileContent, instead of backend.NoConditions{}.

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -115,6 +115,16 @@ func (c generationCondition) ConditionsMet(activeGeneration int64) bool {
 	return true
 }
 
+// resumableUploadEntry holds the in-progress object for a resumable upload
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thank you!

PR #2261, opened 2026-06-27
feat: implement bucket ACL endpoints

gaul:
The JSON API only served object ACLs (/b/{bucket}/o/{object}/acl); bucket ACLs (/b/{bucket}/acl) returned 404, so clients that set or read a bucket's ACL -- e.g. granting allUsers READER for public access -- silently no-op'd.

Store the ACL on the bucket (backend Bucket/BucketAttrs, persisted by both the memory and filesystem backends via a new UpdateBucketACL) and serve the bucketAccessControls endpoints: list/insert on /b/{bucket}/acl and get/update/delete on /b/{bucket}/acl/{entity}.


gaul:
Spurious CI failure:
> Error:  [launcher] xsbt.boot.internal.shaded.coursier.error.ResolutionError$CantDownloadModule: Error downloading org.scala-sbt:sbt:1.6.1

fsouza (approved):
Thanks for the fix on this. Two minor/non-blocking things that would be good to address in a follow-up if you have the time:

1. fs backend wipes ACLs on PATCH (`storageFS.UpdateBucket` always overwrites the metadata)
2. ListBuckets drops the ACL on fs backend

Nothing blocking, so let's ship!

PR #2283, opened 2026-07-16
fix: validate declared MD5/CRC32C checksums on upload

gaul:
fake-gcs-server computed an object's MD5/CRC32C from the uploaded bytes but never compared them against a checksum declared by the client, so an upload whose md5Hash/crc32c did not match the data was silently accepted. Real GCS rejects such uploads with HTTP 400 and reason "invalid".

Parse the declared md5Hash/crc32c from the upload metadata and reject mismatches with 400 on the multipart and resumable upload paths (the single-request media/form paths stream their body and are left as-is).


fsouza (approved):
Thank you!

PR #2304, opened 2026-07-28
Honor response-* query parameters in object downloads

Xepe:
Real GCS allows download URLs to override response headers via `response-*` query parameters (e.g. `response-content disposition`, which signed URLs use to trigger attachment downloads). fake-gcs-server ignored these, serving headers from stored object metadata only — so files that download against real GCS render inline against the emulator.

This change applies the six documented overrides (`response-content-disposition`, `response-content-type`, `response-cache-control`, `response-content-encoding`, `response-content-language`, `response-expires`) from the query string, taking precedence over stored metadata.

Ref: https://cloud.google.com/storage/docs/xml-api/reference-headers#query

Added `TestServerClientObjectReadResponseHeaderOverrides` covering the no-override and override cases

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -1131,12 +1131,8 @@ func (s *Server) downloadObject(w http.ResponseWriter, r *http.Request) {
 		// response headers, see
 		// https://cloud.google.com/storage/docs/xml-api/reference-headers#query
 		for param, header := range map[string]string{
-			"response-cache-control":       cacheControlHeader,
 			"response-content-disposition": contentDispositionHeader,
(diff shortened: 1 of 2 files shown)


fsouza (commented):
Thanks for contributing. This looks good, but the GCS docs only mention `response-content-disposition` and `response-content-type` (though the others are in the S3 api). Can you confirm whether this is undocumented behavior from GCS or if we should remove the extra entries?

Xepe:
@fsouza Good catch on that. I'll remove the extra entries since they are not in the official documentation. 

Thanks.

fsouza (approved):
Thank you!

PR #2308, opened 2026-07-30
fix: honour ifGenerationMatch precondition in simple uploads

thuibr:
simpleUpload passed backend.NoConditions{} to createObject unconditionally, ignoring the ifGenerationMatch query parameter. This caused conditional writes (e.g. optimistic concurrency on manifest files) to always succeed regardless of the supplied generation, making 412 responses unreachable via uploadType=media.

Fix: call wrapUploadPreconditions in simpleUpload and pass the resulting conditions to createObject, consistent with the multipart and resumable upload paths.

Adds TestServerClientSimpleUploadIfGenerationMatch covering gen=0 on new object, gen=0 on existing object, wrong generation, and correct generation.

### DIFF
--- a/fakestorage/upload.go
+++ b/fakestorage/upload.go
@@ -463,6 +463,11 @@ func (s *Server) simpleUpload(bucketName string, r *http.Request) jsonResponse {
 		}
 	}
 
+	conditions, err := s.wrapUploadPreconditions(r, bucketName, name)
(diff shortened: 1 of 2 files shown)


fsouza (approved):

fsouza:
Thanks for contributing!

PR #2309, opened 2026-07-31
feat: honor predefinedAcl and acl on bucket creation

gaul:
buckets.insert accepts a predefinedAcl query parameter and an acl list in the request body, and the real API applies either to the new bucket. The handler decoded only name, versioning and defaultEventBasedHold, so both were accepted and silently dropped: a client asking for a publicRead bucket got a private one and no error, and had to follow up with a separate call to /b/{bucket}/acl to get the access it had already requested.

Nothing else was missing -- BucketAttrs already carries ACL, the backend already stores it, and the standalone bucket ACL endpoints already read and write it. Only the create path ignored the request.

Decode acl into the existing aclRule type, which already unmarshals the entity and role field names, and expand predefinedAcl for the values that name a public or authenticated scope. private and projectPrivate map to no rules, matching what the emulator already produces for a bucket created without an ACL, so the default is unchanged. The project-scoped roles the real API adds are omitted because the emulator does not model projects.

Reject a request carrying both predefinedAcl and acl, and one naming an unknown predefinedAcl, rather than silently preferring one.

Objects already had this: upload.go reads predefinedAcl on every upload path and maps it through getObjectACL. Buckets were the gap.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -89,6 +89,42 @@ func (s *Server) CreateBucketWithOpts(opts CreateBucketOpts) {
 	}
 }
 
+// bucketACLFromRequest resolves the ACL a bucket is created with. The
+// predefinedAcl query parameter and an explicit acl list are mutually
+// exclusive, as they are in the real API.
+func bucketACLFromRequest(predefinedACL string, acl []aclRule) ([]storage.ACLRule, error) {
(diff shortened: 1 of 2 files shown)


fsouza (approved):
Thank you!

gaul:
@fsouza this PR is approved but unmerged -- any work on my side here?

fsouza:
My bad, I missed it

PR #2310, opened 2026-07-31
feat: honor destinationPredefinedAcl on object copy and compose

gaul:
objects.copy, objects.rewrite and objects.compose all accept a destinationPredefinedAcl, and the real API applies it to the object they produce.  Neither handler read it: a rewrite gave the destination the source object's ACL, and a compose left whatever CreateObject assigns, so a client asking for a public destination got a private one and no error.

Take it from the query string in both, mapping it through the existing getObjectACL, which is what the upload paths already do for predefinedAcl.  A rewrite with no destination ACL keeps inheriting the source's, and a compose without one keeps whatever it had, so the default in each is unchanged.

Composing needs the ACL at the point the destination is built, so it joins the attributes ComposeObject already carries.  The gRPC compose passes nil: it forwards none of the destination attributes today, and wiring only the ACL would leave it the odd one out.

Objects uploaded through /upload already had this; copying and composing them did not.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -979,11 +979,19 @@ func (s *Server) rewriteObject(r *http.Request) jsonResponse {
 	if _, err := s.backend.GetBucket(dstBucket); err != nil {
 		return jsonResponse{status: http.StatusNotFound}
 	}
+	// A destination ACL named on the request replaces the one inherited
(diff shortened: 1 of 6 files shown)


fsouza (approved):
Thanks!

PR #2322, opened 2026-08-17
Support deleting a specific object generation

gaul:
DELETE on an object silently ignored the generation query parameter: it archived or removed the live object whatever generation the caller named, and a generation that was already archived could not be removed at all.  Parse the parameter and add DeleteObjectWithGeneration to the backend interface with Cloud Storage's semantics: deleting a named generation removes it outright, live or archived, even on a versioning-enabled bucket, where only a plain delete archives.


fsouza (approved):
Thank you!

PR #2333, opened 2026-08-22
Support pagination when listing buckets

gaul:
Listing buckets ignored maxResults, pageToken and prefix and answered in backend order: every bucket in one response, never a nextPageToken, so a paging client saw either everything at once or, with a real page size, results it had no way to continue.  Parse the three parameters and answer the way the object listing already does: buckets sort by name, the prefix filters, and when more remain past maxResults the nextPageToken names the first bucket of the next page.

### DIFF
--- a/fakestorage/bucket.go
+++ b/fakestorage/bucket.go
@@ -11,6 +11,9 @@ import (
 	"io"
 	"net/http"
 	"regexp"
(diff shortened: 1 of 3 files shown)


fsouza (approved):
Thank you!

PR #2343, opened 2026-08-25
Honor object preconditions outside the upload path

gaul:
The ifGenerationMatch family was answered only where an upload creates an object.  Delete, patch, update, rewrite, copy and compose parsed no preconditions at all, so a caller guarding a mutation against the generation it had just read was told the mutation succeeded whatever the object had become in between -- silently, since a condition the server never reads cannot fail.  Real GCS answers 412 to every one of them.

Move the parsing out of the upload path into a preconditions type that reads the whole family, ifMetagenerationMatch included, and the source family a rewrite names for the object it reads rather than the one it writes.  A rewrite's destination conditions go on to the backend the way an upload's already do, so they are answered where the object is published; the rest are answered against the live object the request is about to change.  Metageneration is compared against the 1 that every object here reports, since nothing revises metadata without minting a new object.

### DIFF
--- a/fakestorage/object.go
+++ b/fakestorage/object.go
@@ -808,12 +808,16 @@ func (s *Server) deleteObject(r *http.Request) jsonResponse {
 	// Calling Close before checking err is okay on objects, and the object
 	// may need to be closed whether or not there's an error.
 	defer obj.Close() //lint:ignore SA5001 // see above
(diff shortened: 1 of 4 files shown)


fsouza (approved):
Thanks for contributing!

PR #2346, opened 2026-08-25
Support gRPC over the plaintext (-scheme http) listener

### LINKED ISSUE #2345: gRPC requests fail for http scheme
## Problem
Running fake-gcs-server with `-scheme http` (or the `-port-http` listener under `-scheme both`) breaks gRPC — clients get no usable response, e.g.:
```
docker run -d -p 4443:4443 fsouza/fake-gcs-server -scheme http
grpcurl -plaintext localhost:4443 google.storage.v1.Storage.ListBuckets
```
Server-side log for that request:
```
time=2026-08-25T15:17:53.529Z level=INFO msg="172.19.0.1 - - [25/Aug/2026:15:17:53 +0000] \"PRI * HTTP/2.0\" 404 19\n"
```

## Why
`http.Serve` (stdlib `net/http`) only ever speaks HTTP/1.1 on a plain TCP listener; HTTP/2 is only enabled via TLS+ALPN, which only happens in `createListener`'s `https` branch. gRPC requires HTTP/2, so on the plaintext listener the server can't recognize the HTTP/2 connection preface (`PRI * HTTP/2.0`) — it gets parsed as a bogus HTTP/1.1 request line (`PRI` as method, `*` as path), hence the 404 above.

## Why this matters
Client libraries that support plaintext gRPC (e.g. the Java `google-cloud-storage` client, via `StorageOptions.grpc().setHost(...)`) only select plaintext transport when the endpoint scheme is `http` — there's no supported way in that client to skip TLS certificate validation for `https`. Using gRPC against a local emulator over `https` means dealing with self-signed-cert trust (generating certs, importing into a truststore, etc.) just for local/dev use. Plaintext `-scheme http` is the way to avoid that complexity, but gRPC doesn't work there currently.

---
*This issue description was drafted with AI assistance (Claude).*

altaiezior:
Fixes #2345

## Problem
Running with `-scheme http` (or the `-port-http` listener under `-scheme both`) breaks gRPC — see #2345 for the full root-cause writeup.

## Fix
Wraps the plaintext listener's routing handler with `golang.org/x/net/http2/h2c`, adding HTTP/2-cleartext support while still transparently serving HTTP/1.1 (the JSON API) unchanged. The TLS listener is untouched — it already gets HTTP/2 via ALPN. `golang.org/x/net` was already an indirect dependency (via grpc/otel); this promotes it to direct.

No test added — this code path (`startServer`'s scheme/transport handling) has no existing test coverage in the repo today, so this change doesn't introduce a new gap.

## Testing
- `go build ./...`, `go vet ./...`, `go test -race -vet all -mod readonly ./...` all pass.
- Manually verified `grpcurl -plaintext` against the built binary with `-scheme http`.

---
*This PR description was drafted with AI assistance (Claude).*

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -71,7 +71,7 @@ require (
 	github.com/googleapis/enterprise-certificate-proxy v0.3.20 // indirect
 	github.com/googleapis/gax-go/v2 v2.23.0 // indirect
 	go.opencensus.io v0.24.0 // indirect
-	golang.org/x/net v0.58.0 // indirect
+	golang.org/x/net v0.58.0
 	golang.org/x/sync v0.22.0 // indirect
 	golang.org/x/sys v0.47.0 // indirect
 	golang.org/x/text v0.41.0 // indirect
--- a/main.go
+++ b/main.go
@@ -22,6 +22,8 @@ import (
 	"github.com/fsouza/fake-gcs-server/fakestorage"
 	"github.com/fsouza/fake-gcs-server/internal/config"
 	"github.com/fsouza/fake-gcs-server/internal/grpc"
+	"golang.org/x/net/http2"
+	"golang.org/x/net/http2/h2c"
 )
 
 func createListener(logger *slog.Logger, cfg *config.Config, scheme string) (net.Listener, *fakestorage.Options) {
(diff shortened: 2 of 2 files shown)


altaiezior:
@fsouza the changes are done and I have also tested them loocally, would you be able to look into this issue. Thanks

